PR #7, opened 2017-03-07
Fix trace of p.bclr and p.bset instructions

asiz:
When these instructions were implemented in e423b2c197dbed08968a6ddcf8356388e07b5f30, their effective `funct3` fields were accidentally switched. This was partially fixed in 7e07801be183f665e382a0667599ec34a9fc484b, but the tracer still had them wrong. This single commit PR fixes that.

By the way, I noted that these instructions are not documented in the datasheet under `docs`. The online [User Manual](http://www.pulp-platform.org/wp-content/uploads/2017/02/ri5cy_user_manual.pdf) has an additional chapter (Chapter 13: Instruction Set Extensions). Shouldn't that be included here?


davideschiavone:
this has been fixed in the pulpinov1 branch.
thanks for contributing!
Davide

PR #44, opened 2018-07-05
Fix Verilator model build

gmarkall:
The most recent commit add a new file - `../rtl/register_file_test_wrap.sv` - which needs to be Verilated too.

This fixes the build - the testbench produces the expected output once this fix is applied.


davideschiavone:
Thanks a lot :) 👍 

Merging now :)

Issue #77, opened 2019-06-11
Typo in riscv_pmp.sv


zdogs:
“ifdef EN_NAPOT_RULE_64KBB”    should be “ifdef EN_NAPOT_RULE_64KB” line  360

bluewww:
Thanks. Line is here https://github.com/pulp-platform/riscv/blob/6898bd17f608f9d620d0335c933b62e3e24a8bfc/rtl/riscv_pmp.sv#L360

PR #78, opened 2019-06-11
Fix typo line 360

zdogs:
Hi please find fix of riscv_pmp.sv described in issue #77


bluewww:
We need to find out if we use the old macro name somewhere (pulpissimo etc.) to not break anything

davideschiavone:
thansk a lot

PR #117, opened 2019-08-15
Add `fpnew` as git submodule

sabhiram:
Hey folks, thanks for all the work on this. 

We currently do some systemc work that requires a riscv model which this repo provides nicely. In trying to automate some of our build-flow we ran across this somewhat icky `git clone` being done from a Makefile inside a source directory. 

I took the liberty to clean this up somewhat in our fork - please review if you wish, and do with this PR as you see fit :)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "tb/core/fpnew"]
+	path = tb/core/fpnew
+	url = https://github.com/pulp-platform/fpnew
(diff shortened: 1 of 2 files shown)


bluewww:
I really don't like adding git submodules. I added the same makefile dependency thing to the `verilator-model` testbench among other fixes, see #146. The ideal solution from my point of view would be to get [bender](https://github.com/fabianschuiki/bender) to work.

PR #134, opened 2019-08-28
Add misa

bluewww:
Previously we just set this register to zero, which is allowed by the
spec, but I think at this point we might as well implement it.


davideschiavone:
Should we use the PULP_SECURE parameter for the User Mode? Why is it commented?

bluewww:
Do we have full User Mode support? I can put it in.

bluewww:
Ok I put it in, I'll merge

PR #181, opened 2019-10-08
:bug: Make pipeline stall on all CSR accesses (#169)

### LINKED ISSUE #169: Illegal Instruction Exception not Raised - Dynamic Rounding Mode
**RISC-V Specification:**

- _"Floating-point operations use either a static rounding mode encoded in the instruction, or a dynamic rounding mode held in frm."_ 
- _"Rounding modes are encoded as shown in Table 11.1. A value of 111 in the instruction’s rm field selects the dynamic rounding mode held in frm."_
- _"If frm is set to an invalid value (101–111), any subsequent attempt to execute a floating-point operation with a dynamic rounding mode will raise an illegal instruction exception."_

**Issue Description:**

An illegal instruction exception is not raised for the case when a floating-point instruction uses a dynamic rounding mode while the frm is set to an invalid value. 

**Example:**

As shown below, the instruction 32'hc011f1d3 (fcvt.wu.s x3, f3) is decoded at time point t##0 where rm field has the value 111, and frm has the value 101 (written by a previous instruction), with no illegal instruction being flagged, as illegal_insn_dec is de-asserted.

issue_8

---

Product: OneSpin 360 DV-Verify
App: RVV
Tool's version: 2019.2.2

stmach:
Fixes bugs where changed CSR state is not visible to following instructions such as when changing the FP status CSR.
Fixes #169.

### DIFF
--- a/rtl/riscv_decoder.sv
+++ b/rtl/riscv_decoder.sv
@@ -2352,10 +2352,8 @@ module riscv_decoder
           end
 
           if(~csr_illegal)
-            if (instr_rdata_i[31:20] == 12'h300 || instr_rdata_i[31:20] == 12'h000  || instr_rdata_i[31:20] == 12'h041 ||
-              instr_rdata_i[31:20] == 12'h7b0 || instr_rdata_i[31:20] == 12'h7b1 || instr_rdata_i[31:20] == 12'h7b2 || instr_rdata_i[31:20] == 12'h7b3) //debug registers
-              //access to xstatus
-              csr_status_o = 1'b1;
+            // Just stall the pipeline on all CSR accesses
+            csr_status_o = 1'b1;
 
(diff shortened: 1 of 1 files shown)


davideschiavone:
Shall we merge @stmach ?

stmach:
As in the current version of the core state-modifying CSRs have to be explicitly black-listed, this fix is kind of a "shotgun" approach to missing some registers - and hence loses performance.
A cleaner (and safer) approach would be to stall it by default (like this PR) but have a white-list of safe CSRs that don't need to flush the pipeline. We can merge this and think of opening a new issue for this as performance improvement.

Btw, the signal name `csr_status_o` that causes the pipeline to be flushed on CSR access is a bit misleading, consider refactoring.

PR #200, opened 2019-11-28
Pull request from OpenHW Group

MikeOpenHWGroup:
Hi, this is Mike Thompson from the OpenHW Group.  We are doing some preliminary work to prepare for the CORE-V CV32 project, which will be based on the RI5CY core.   I've done some work which I think may be of benefit to RI5CY and hence this pull-request.

The big changes in this PR:

1. Add Makefile rules for the Metrics SystemVerilog simulator _**dsim**_.  These rules will not affect anyone not using dsim.
2. Separate the RTL and TB source files.  This involved significant changes to Makefile variables and also moving one file from the tb/core directory to the rtl directory.
3. Removed the rvc.S testcase as it does not compile (we have created an issue to ensure it does not get forgotten).

One minor issue - my updates to README.md are relevant to my fork only - you probably don't want to merge that.


bluewww:
Hi Mike, thanks for your PR.
A few comments:
* `rvc.x` is imported from [riscv-tests](https://github.com/riscv/riscv-tests). If I recall correctly this test caused issues with different compiler versions at some point.
* CI seems to be broken from a commit between 43375ff and c6dd65d. I couldn't figure out on the fly why.
* Support for more simulators is always welcome. I don't mind the `Makefile` changes.

bluewww:
There is no point in adding `fpnew_pkg.sv` to this repository. This file doesn't belong here. RI5CY depends on fpnew so its much better to just pull the dependency instead of having ltwo `fpnew_pkg.sv` in a system which could cause name conflicts and go out of sync.

MikeOpenHWGroup:
As far as I could tell, the only source from fpnew used by riscy is the package file (specifically, a set of parameters).   Its not clear why riscy would create a dependency on another project just for a simple package file.  Am I missing something?

bluewww:
This branch https://github.com/pulp-platform/riscv/tree/enable-fp-tb uses RI5CY with FPU capabilities, we haven't cleaned up and merged it yet.

MikeOpenHWGroup:
Ah, Maintenant, je comprends.

OK, I'll fix my branch to include all code for fpnew.   Can we leave it under the 'rtl' directory?  It was pulled into the 'tb' directory.  As much as possible, it is nice to have a clean separation of RTL and Testbench files.

Please reject this pull-request and I'll create another when I fix the branch at my end.

Thanks for your support!

bluewww:
You can just force push on your branch anways. Its a development branch, you can do whatever. Infact this is the usual way to do it.

Silabs-ArjanB:
Hi @MikeOpenHWGroup Unfortunately this PR has been laying around for a very long (much too long) time and it is now completely outdated. Could you please either remove the pull request or issue a new one with changes that are still relevant?

MikeOpenHWGroup:
> this PR has been laying around for a very long (much too long) time and it is now completely outdated.

Agreed.  Way back in November of 2019 I recommended killing this PR.

> Could you please either remove the pull request or issue a new one with changes that are still relevant?

I will close this PR.

PR #202, opened 2019-12-04
A Extension: Decode ops and send them to memory as `data_atop_o`

accuminium:
This implements RISC-V's 'A' standard extension for atomic instructions by decoding atomic operations and feeding them through to a new `data_atop_o` port.  The tracer has been updated as well.

### DIFF
--- a/rtl/include/riscv_defines.sv
+++ b/rtl/include/riscv_defines.sv
@@ -52,6 +52,7 @@ parameter OPCODE_OP_FMSUB  = 7'h47;
 parameter OPCODE_OP_FNMSUB = 7'h4b;
 parameter OPCODE_STORE_FP  = 7'h27;
 parameter OPCODE_LOAD_FP   = 7'h07;
+parameter OPCODE_AMO       = 7'h2F;
 
 // those opcodes are now used for PULP custom instructions
 // parameter OPCODE_CUST0     = 7'h0b
--- a/rtl/include/riscv_defines.sv
+++ b/rtl/include/riscv_defines.sv
@@ -302,6 +302,18 @@ parameter JT_JAL  = 2'b01;
 parameter JT_JALR = 2'b10;
 parameter JT_COND = 2'b11;
 
+// Atomic operations
+parameter AMO_LR   = 5'b00010;
+parameter AMO_SC   = 5'b00011;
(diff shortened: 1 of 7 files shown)


accuminium:
Not all memory systems implement atomic operations, and with this PR the core relies on the remote execution of the AMO but has no way to get a positive or negative acknowledgement (RISC-V AMOs cannot fail).  Thus, RI5CY would happily execute AMOs even if the memory system does not implement them, with could be dangerous.

To solve this problem, I suggest to add a `parameter bit A_EXTENSION` to `riscv_core`, `riscv_id_stage`, and `riscv_decoder`.  The decoder would only decode instructions from the A extension if this bit was set, and otherwise raise an invalid instruction.  This way, the `data_atop_o` can be safely left unconnected if `A_EXTENSION = 0`.

davideschiavone:
Hello Andres, thanks a lot for the PR.

Can you please add this parameter you are mentionig to avoid the encoding of RVA instructions in case they are not used?

bluewww:
Testbench extended in https://github.com/pulp-platform/riscv/pull/204. That PR goes after this one.

* Is it correct that `data_we_o` is kept high once the atomic instructions begin executing? I had to slightly adapt `amo_shim.sv`.
* What is required to support lrsc, the `amo_shim` doens't have support for it?

accuminium:
> Can you please add this parameter you are mentionig to avoid the encoding of RVA instructions in case they are not used?

Yes, will do.

> * Is it correct that `data_we_o` is kept high once the atomic instructions begin executing? I had to slightly adapt `amo_shim.sv`.

`data_we_o` continues to be driven by the EX stage (), so I think this should not differ from other writes. The decoder sets its `data_we_o` when decoding AMOs.

> * What is required to support lrsc in the `amo_shim`?

In a testbench with a single-threaded single core, you have to add one register that holds a reservation on an address.  That register is set by an LR and checked and cleared by an SC.  This logic can also be implemented outside the `amo_shim` in the TB.  Beyond this simple TB case, when many harts can access the `amo_shim`, harts need to identify themselves with a unique ID, you need one reservation register per hart, and you need to invalidate all reservations to an address upon any write.

accuminium:
I have added the parameter as discussed.

PR #204, opened 2019-12-09
Extend core tb to support atomics

bluewww:
Features
* Enable `riscv-tests`'s `rv32ua`
* Disable `lrsc` test (not implemented)
* Add atomic shim (`amo_ infront of RAM model and wire up atop. Translate meaning of `data_atop_i` to match the one from the shim
* Patch write enable behavior in `amo_shim.sv` to work with RI5CY.

TODO
* Support random stall generator (currently only verilator tested)


davideschiavone:
can this be merged now @bluewww ? I would like to merge the RVA from @accuminium into the core and have the TB aligned

bluewww:
we can merge after https://github.com/pulp-platform/riscv/pull/202 goes in.

Issue #221, opened 2020-01-31
SCM-based register file is not cleared at reset


FrancescoConti:
This causes RTL vs gate-level simulation mismatch. @davideschiavone I can submit a PR with a fix for this (it's the same I already implemented e.g. in https://github.com/pulp-platform/hwpe-ctrl ). Would you consider merging it?

davideschiavone:
really ? I never had problem with post layout simulations. What kind of problem is that ? can you describe it a bit more? of course I consider your PR to be merged :)

FrancescoConti:
It's not exactly a bug, but in some cases to make PS simulations work I had to manually clear the register file in Modelsim with `force -deposit regfile/Q 1'h0`. It's a bit annoying :)
Right now I'm doing that on a synthesis of a somewhat recent version of PULPissimo, the testbench dies miserably when the debug module tries to halt the core.

bluewww:
But doesn't that mean it's an issue with the testbench? I think its a quite common assumption that cpu registers are unitialized/hold random garbage on reset.

FrancescoConti:
@bluewww this is *mostly* a testbench related issue . But actually if you implement the register-file with flip-flops you will get a subtly different state, when you assume that the state is identical, leading to divergence with expected behavior in case of buggy code (I actually saw this on FPGA with a much older version of Riscy :) )

gautschimi:
you should clear all relevant registers in the boot sequence. or better write code that does not depend on uninitialized registers:)

davideschiavone:
Yes, in my opinion this comes from buggy code - as you should not use a register content if not utilized.

However, as @FrancescoConti wrote in the PR, resettable latches are more expensive and not always available in libraries. The cost of this change looks negligible (1FF and 33 OR2 gates).

 I do not have a strong opinion neither to accept nor to deny this PR - I can actually forwards the question to openHW

FrancescoConti:
Absolutely, no strong opinion on my side either -- just a proposal as I think this is the kind of subtle difference that it's better to avoid at the root.

davideschiavone:
What I like about accepting it is that the 2 cores (FF or Latches) would have the same state.
Also it would save some instructions.
The cons is that actually it should be a SW task.

MikeOpenHWGroup:
In industry it is common practise to reset everything, unless it becomes impractical in physical design (all those resets create a lot of loads).  The idea is to protect software from itself.

[...]

PR #222, opened 2020-02-04
Clearing the content of latch-based regfile at boot.

FrancescoConti:
This PR addresses issue #221 .

Adding a flip-flop generating a 1-cycle clear strobe when it samples
the first two clock edges after reset. In this way, the content of
the register-file is cleared in a similar way to what happens in the
FF-based register-file.


davideschiavone:
Hi @FrancescoConti 

Thanks a lot for your PR.

Can you describe a little bit more why the RESET is not zeroing the latches?
Also, do you know the impact in area/performance that this extra "if elseif" that has been written to handle the clear signal from the flip-flop has on the core?

Thanks

FrancescoConti:
The reset is not zeroing the latches, nor has it ever in SCMs :) about the reason: typically this is because we deployed SCM's as memory replacements (non-reset). Moreover, at least in the first standard cell library I could check, resettable latches are ~15% larger than the correspondent non-resettable ones. Also, I am unsure whether all libraries include resettable latches at all (maybe someone else can confirm / deny).

Rgd the area cost of this change: as far as I can count, the additional cost is 1 flip-flop, plus 33 OR2 gates being transformed in OR3 (for a 32-words register file).

davideschiavone:
Ok I had misunderstand the issue - this looks like SW issue solved in HW. 
However as I wrote in the related issue, it looks anyway a convenient feature.
Forwarding this to OHW

FrancescoConti:
The CI is doing something funny. Between a failing and a successful run I literally just removed a newline 🤣

davideschiavone:
@FrancescoConti yes this happened to me as well....

PR #224, opened 2020-02-07
Added retire event. Moved insn_disas, insn_pc, and insn_val assignmen…

GTumbush:
This pull request is to address https://github.com/pulp-platform/riscv/issues/218
1) Added a retire event.
2) Moved assignment to insn_disas, insn_pc, and insn_val assignment to after retire event trigger so these variables change only when an instruction is retired.  This enables a testbench to wait on the retire event and collect the results of the retired instructions.  If you want to keep the assignment of these 3 variables as-is I can create separate signals.

### DIFF
--- a/rtl/riscv_tracer.sv
+++ b/rtl/riscv_tracer.sv
@@ -104,6 +104,8 @@ module riscv_tracer (
   integer      cycles;
   logic [ 5:0] rd, rs1, rs2, rs3, rs4;
 
+  event        retire;
+  
   typedef struct {
     logic [ 5:0] addr;
     logic [31:0] value;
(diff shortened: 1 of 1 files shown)


GTumbush:
FYI, the checks failed because of:


Job                                     | OS     | State
--                                       | --      | --
214.1 download pulp gcc | Linux | errored

I don't think this has anything to do with the changes I made to riscv_tracer.sv

davideschiavone:
Thanks @GTumbush , for me it is ok to push. As new rule in OHW, I want to reviewrs for every PRs (4 eyes better than 2).
So one is me, the other can come from the Core Team. Can I ask you @sjthales to review the change as well?

davideschiavone:
Also, @GTumbush , be sure to have a valid Eclipse Foundation account to make pushes to the openhwgroup repositories.

sjthales:
review_#224
@GTumbush, @davideschiavone
I ran firmware-vsim- run with riscv_tracer.sv modified by GTumbush, I  noticed a weird behavior with a compressed instruction.
For the instruction below
    90de:	478d                	li	a5,3
the fetch instruction is coherent but when there is a retire event on this instruction, we have:
insn_pc = 0x000090dE
insn_disas = addi x15, x0, 3
insn_val = 0x00300793

is that normal?

davideschiavone:
Hi @sjthales , thanks for your quick reply!
Yes it's normal as the tracer dumps the uncompressed version of the instruction

sjthales:
ok, if it's normal, so that's ok for me.

PR #151, opened 2019-09-05
Remap CSR registers

bluewww:
# Changes
* Fix hwloop CSRs in manual
* Add misa to manual
* Add mscratch to manual
* Delete from pcer/pcmr custom read/write user-mode space
* Map hwloop registers to user-mode CSRs (0x8C0-0x8C6)
* TODO: map readable pcer registers to user-mode readable CSRs
* TODO: consider moving fprec, uhartid, upriv

Fixes #142 
# Overview of CSR space
Current state of CSR mapping in RI5CY.
Variables between earmuffs are custom CSRs as of now.
Variables between slashes are new mappings.

    USER SPACE
    |  addr | name      | priv | use according | note                        |
    |       |           |      | to spec       |                             |
    |-------+-----------+------+---------------+-----------------------------|
    | 0x000 | ustatus   | rw   | standard rw   |                             |
    |-------+-----------+------+---------------+-----------------------------|
    | 0x001 | fflags    | rw   | standard rw   |                             |
    | 0x002 | frm       | rw   | standard rw   |                             |
    | 0x003 | fcsr      | rw   | standard rw   |                             |
    |-------+-----------+------+---------------+-----------------------------|
    | 0x005 | utvec     | rw   | standard rw   |                             |
    | 0x006 | *fprec*   | rw   | standard rw   | [ ] *move*, reserved space  |
    |-------+-----------+------+---------------+-----------------------------|
    | 0x014 | *uhartid* | r    | standard rw   | [ ] *move*, reserved space  |
    | 0x041 | uepc      | rw   | standard rw   |                             |
    | 0x042 | ucause    | rw   | standard rw   |                             |
    | 0xc10 | *upriv*   | r    | standard rw   | [ ] *move*, collides        |
    |-------+-----------+------+---------------+-----------------------------|
    | 0x8c0 | /hwlp0/   | rw   | custom rw     | [x] map hwloop              |
    |   ... |           |      |               |                             |
    | 0x8c6 | /hwlp6/   | rw   | custom rw     |                             |
    |-------+-----------+------+---------------+-----------------------------|
    | 0xcd0 | /pccrs/   | r    | custom r      | [ ] map performance counter |
    |   ... | ...       |      |               |                             |
    | 0xce1 | /pccrs/   | r    | custom r      |                             |
    |-------+-----------+------+---------------+-----------------------------|

    MACHINE SPACE
    |  addr | name      | priv | use according | note                                |
    |       |           |      | to spec       |                                     |
[...]

PR #226, opened 2020-02-10
CLINT and CLINTx extension

davideschiavone:
**CLINT**

-  RI5CY is now compatible with the RISC-V Clint.
 
- Code heavily based on the Ibex Core from lowRISC

- See documentation at https://ibex-core.readthedocs.io/en/latest/exception_interrupts.html

**CLINTx**

- Added 32 interrupt lines more (namely FAST 32 to 63)

-  Added MTVECx and MIEx to control where the core jumps on these 32 requests and to control the interrupt enables

-  ACK signal from the core now active only on these 32 interrupt requests

 
**TestBench**

- Adapted firmwares and linker scripts to have 32 lines more in the IVT

- Added interrupt test

**TODO**

-  Update documentation

- Adapt DM testbench

-


rtl/include/riscv_defines.sv
```diff
@@ -195,12 +195,80 @@ parameter VEC_MODE8  = 2'b11;
...
+  // Machine information
+  CSR_MHARTID   = 12'hF14,
+
+  // Machine trap setup
+  CSR_MSTATUS   = 12'h300,
+  CSR_MISA      = 12'h301,
+  CSR_MIE       = 12'h304,
+  CSR_MTVEC     = 12'h305,
+  CSR_MIEX      = 12'h306,
+  CSR_MTVECX    = 12'h307,
```
vogelpi: Sorry @davideschiavone, @bluewww and @haugoug for being late. While adopting the additional 32 fast interrupts in Ibex (for PULPissimo) I came across this PR. Your current mapping of the CLINTx CSRs will for sure cause problems soon:
1. CSR_MIEX is overlapping with the `mcounteren` CSR defined in the spec.
2. CSR_MTVEX and CSR_MIPX might be used for other CSRs introduced in future versions of the spec.

I recommend to move all CSRs related to CLINTx into on of the *custom* CSR ranges defined in the spec, e.g., 0x7C0 - 0x7FF or 0xBC0-0xBFF (Table 2.1 of the RISC-V Priv. Spec.). And I recommend to do this now before people start to write software for it.

Please let me know where you will put it. I would like to align the Ibex PULPissimo branch.

davideschiavone: Thanks @vogelpi! This is indeed a issue to solve asap, any recommendation @haugoug , @Silabs-ArjanB , @bluewww  ?

haugoug:
Yes this should be moved to the custom area

rtl/include/riscv_defines.sv
vogelpi: On Ibex, we do not use any custom/non-standard CSRs so far. So what ever you choose, I should be able to align the PULPissimo branch of Ibex. Just let me know what you end up with ;-)

bluewww:
There are more CRS collisions. See https://github.com/openhwgroup/cv32e40p/pull/151. Maybe its worth going through this PR.

rtl/include/riscv_defines.sv
Silabs-ArjanB: > 
> 
> Thanks @vogelpi! This is indeed a issue to solve asap, any recommendation @haugoug , @Silabs-ArjanB , @bluewww ?

Anywhere in the appropriate custom CSR range (as per Table 2.1 of the RISC-V Priv. Spec) would be fine.

vogelpi: @davideschiavone and @Silabs-ArjanB, any news from your side? Did you decide on a set of CSR addresses for these custom CSRs?

davideschiavone: I suggest `0x7D6` and  `0x7D7`  instead of `0x306` and `0x307`.

Can you open an issue @vogelpi? @Silabs-ArjanB what do you think?

Silabs-ArjanB: We need to move MIPX as well. Unfortunately it will not be possible to have MIEX, MTVECX, MIPX spaced out in the same way as MIE, MTVEC, MIP (as there simply isn't enough space in the CSR custom R/W spaces), otherwise I would have preferred address ending in 04, 04, 44 respectively. Now my proposal would be to simply start at 0x7d0, so MIEX = 0x7d0, MTVECX = 0x7d1, MIPX = 0x7d2

davideschiavone: 7D0, 7D1 AND 7D2 are fine for me. Would you prepare an PR @Silabs-ArjanB ? I can assign the issue to you

vogelpi: Thanks @davideschiavone for creating the issue here: #253 . Let's hope it will be solved at some point.

Issue #227, opened 2020-02-11
Reads of `mip` CSR should not depend on corresponding fields in `mie`


davideschiavone:
As for Ibex, RI5CY has the issue reported in

https://github.com/lowRISC/ibex/issues/567

davideschiavone:
6a4e57b757fae35c447db20ac2683b1239441631

PR #231, opened 2020-02-12
:fix: for issue #227

davideschiavone:
This PR should solve issue #227

### DIFF
--- a/rtl/riscv_core.sv
+++ b/rtl/riscv_core.sv
@@ -358,6 +358,7 @@ module riscv_core
   logic                             instr_err_pmp;
 
   // interrupt signals
+  logic        irq_en_pending     ;
   logic        irq_pending;
   logic [5:0]  irq_id;
 
@@ -730,7 +731,7 @@ module riscv_core
 
 
     // Interrupt Signals
-    .irq_pending_i                ( irq_pending          ), // incoming interrupts
+    .irq_pending_i                ( irq_en_pending       ), // incoming interrupts
     .irq_id_i                     ( irq_id               ),
     .irq_sec_i                    ( (PULP_SECURE) ? irq_sec_i : 1'b0 ),
     .m_irq_enable_i               ( m_irq_enable         ),
@@ -1015,6 +1016,7 @@ module riscv_core
     .irq_fast_i              ( irq_fast_i         ),
     .irq_nmi_i               ( irq_nmi_i          ),
     .irq_fastx_i             ( irq_fastx_i        ),
+    .irq_en_pending_o        ( irq_en_pending     ),
     .irq_pending_o           ( irq_pending        ), // IRQ to ID/Controller
     .irq_id_o                ( irq_id             ),
     // debug
--- a/rtl/riscv_core.sv
+++ b/rtl/riscv_core.sv
@@ -358,7 +358,6 @@ module riscv_core
   logic                             instr_err_pmp;
 
   // interrupt signals
-  logic        irq_en_pending     ;
   logic        irq_pending;
   logic [5:0]  irq_id;
 
@@ -731,7 +730,7 @@ module riscv_core
 
 
     // Interrupt Signals
-    .irq_pending_i                ( irq_en_pending       ), // incoming interrupts
+    .irq_pending_i                ( irq_pending          ), // incoming interrupts
     .irq_id_i                     ( irq_id               ),
     .irq_sec_i                    ( (PULP_SECURE) ? irq_sec_i : 1'b0 ),
     .m_irq_enable_i               ( m_irq_enable         ),
@@ -1016,7 +1015,6 @@ module riscv_core
     .irq_fast_i              ( irq_fast_i         ),
     .irq_nmi_i               ( irq_nmi_i          ),
     .irq_fastx_i             ( irq_fastx_i        ),
-    .irq_en_pending_o        ( irq_en_pending     ),
     .irq_pending_o           ( irq_pending        ), // IRQ to ID/Controller
     .irq_id_o                ( irq_id             ),
     // debug
--- a/rtl/riscv_cs_registers.sv
+++ b/rtl/riscv_cs_registers.sv
@@ -1035,7 +1035,8 @@ end //PULP_SECURE
   always_comb
   begin
 
-    if      (menipx[31])         irq_id_o = 6'd63;
+    if (menip.irq_nmi)           irq_id_o = 6'd31;
+    else if (menipx[31])         irq_id_o = 6'd63;
     else if (menipx[30])         irq_id_o = 6'd62;
     else if (menipx[29])         irq_id_o = 6'd61;
     else if (menipx[28])         irq_id_o = 6'd60;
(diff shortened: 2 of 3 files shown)


rtl/riscv_core.sv
```diff
@@ -730,7 +731,7 @@ module riscv_core
 
 
     // Interrupt Signals
-    .irq_pending_i                ( irq_pending          ), // incoming interrupts
+    .irq_pending_i                ( irq_en_pending       ), // incoming interrupts
```
Silabs-ArjanB: Please rename the irq_pending_i pin of riscv_id_stage (all the way down in the hierarchy if needed)  to irq_en_pending_i. The current renaming of signals when passing through a hierarchy will become confusing during debug (especially as irq_pending and irq_en_pending have different semantics).

rtl/riscv_core.sv
```diff
@@ -1015,6 +1016,7 @@ module riscv_core
     .irq_fast_i              ( irq_fast_i         ),
     .irq_nmi_i               ( irq_nmi_i          ),
     .irq_fastx_i             ( irq_fastx_i        ),
+    .irq_en_pending_o        ( irq_en_pending     ),
```
Silabs-ArjanB: A comment about the following code line (can't attach it there as that line is not changed):

assign clock_en      = PULP_CLUSTER ? clock_en_i | core_busy_o : irq_pending | debug_req_i | core_busy_o;

I believe that clock_en should depend on irq_en_pending instead of irq_pending. I say that based on that following statement from section "3.2.3 Wait for Interrupt" of "The RISC-V Instruction Set Manual
Volume II: Privileged Architecture":

> The WFI instruction can also be executed when interrupts are disabled. The operation of WFI must
> be unaffected by the global interrupt bits in mstatus (MIE/SIE/UIE) and the delegation registers
> [m|s]ideleg (i.e., the hart must resume if a locally enabled interrupt becomes pending, even if it
> has been delegated to a less-privileged mode), **but should honor the individual interrupt enables
> (e.g, MTIE) (i.e., implementations should avoid resuming the hart if the interrupt is pending but
> not individually enabled)**. WFI is also required to resume execution for locally enabled interrupts
> pending at any privilege level, regardless of the global interrupt enable at each privilege level.
>

rtl/riscv_cs_registers.sv
```diff
@@ -1022,58 +1036,58 @@ end //PULP_SECURE
...
+    else if (menipx[ 8])         irq_id_o = 6'd40;
+    else if (menipx[ 7])         irq_id_o = 6'd39;
+    else if (menipx[ 6])         irq_id_o = 6'd38;
+    else if (menipx[ 5])         irq_id_o = 6'd37;
+    else if (menipx[ 4])         irq_id_o = 6'd36;
+    else if (menipx[ 3])         irq_id_o = 6'd35;
+    else if (menipx[ 2])         irq_id_o = 6'd34;
+    else if (menipx[ 1])         irq_id_o = 6'd33;
+    else if (menipx[ 0])         irq_id_o = 6'd32;
+    else if (menip.irq_nmi)      irq_id_o = 6'd31;
```
Silabs-ArjanB: I wonder if we should make NMI the highest priority interrupt; what do you think? (The priority order in general does not follow the index number, i.e. msi is higher priority than mti)

rtl/riscv_core.sv
davideschiavone: That's indeed different from what lowRISC changed in Ibex. I would ask them how they solved this. thanks for catching it

vogelpi: That's exactly how we implemented it in Ibex: the `irq_pending` in Ibex does NOT factor in the `mstatus.MIE` but it factors in the individual interrupt enables in `mie`.

In Ibex, the `mstatus.MIE` is only factored in inside the controller when deciding whether an interrupt should be handled or not. It is not involved in the WFI instruction at all.

vogelpi: Fore more details see @GregAC's comment here: https://github.com/lowRISC/ibex/issues/567#issuecomment-588359593

rtl/riscv_cs_registers.sv
davideschiavone: I would keep it aligned with Ibex for the moment

rtl/riscv_core.sv
davideschiavone: Ok thanks for the clarifcation guys. I will change this part

rtl/riscv_cs_registers.sv
vogelpi: Yes, this is correct. The general rule is that the highest priority is given to the interrupt with the highest ID. The only exception to the rule is the software interrupt which is given higher priority than the timer interrupt. The reason for this is that they wanted to make software interrupts addressable with the five-bit immediate in CSR instructions (just use a single instruction for reasons of performance). Thus, the software interrupts must be in the lowest four bits of `mip`.

As for the NMI: The NMI is the highest interrupt Ibex supports and also has the highest priority. The plan was to add also the 32 additional interrupts in the `x`-versions of the `mie`/`mip` registers, but just on a branch for compatibilbity with PULP/PULPIssimo. In these systems, the normal `mie`/`mip` including NMI interrupts are not used.

vogelpi: Thinking about this more, you could maybe just give the NMI the highest priority without changing the ID (creating a second exemption to the rule) as @Silabs-ArjanB suggested. In the end, the `x`-versions just give you 32 more fast interrupts, nothing fancy. But I would recommend to document this properly, we had quite some discussions about the NMI.

davideschiavone: I am not 100% in favor to move away from the general rule. Thus I would change the MSI and MTI priority order (done in the last commit @Silabs-ArjanB )

Regarding the NMI I would be in favor in principle, but we added the extra 32 interrupts for keeping compliancy with PULP originally. If we now move the NMI to FASTX63, we will have 32 interrupts jumping to the MTVECX IVT as PULP requested but one is now not maskable.

@haugoug, what do you think?

Silabs-ArjanB: @davideschiavone I was not implying that there was anything wrong with the MSI-MTI priority; those were correct, although it might not 'look' logical at first sight (but this is explained perfectly well in the offical RISC-V spec). My point is just make NMI the highest priority; you do not need to change the id, exactly as @vogelpi writes two comments above this one.

davideschiavone: Ok sorry for the confusion. I overwrote the last commit with the MTI and MSI priority fixed (my bad). For what concern the NMI I would wait for @haugoug, who is expert in SW.

It's true that you don't have to change the id or the MTVEC, but this non linearity looks somehow strange in my opinion. I don't have strong opinions about it, I would wait for SW people inputs.

Silabs-ArjanB: Let's hear from a software person indeed, but it will be non-linear anyway because of the RISC-V spec relateed to MTI/MSI.

haugoug: This non-linearity is indeed annoying but since it's already there, I think it is better to keep the NMI on 31 with an exception to have highest priority so that it is still using MTVEC to not force people to use MTVECx just for NMI.

davideschiavone: Ok let's go for it :+1:

davideschiavone:
@Silabs-ArjanB , please find the last commits with the suggested fixes

Issue #229, opened 2020-02-11
Should we remove the TB directory from this project?


MikeOpenHWGroup:
The OpenHW Group CORE-V projects have separate GitHub projects for design and verification.  As such, the production testbench for cv32e40p is not the one in the `tb` directory.  Given that, I recommend it be removed from this repo altogether and replaced with a pointer to the core-v-verif project.

The obvious objections to this proposal is that the testbench in cv32e40p can be used by the designers for basic sanity.   This is not a good idea because it allows design updates to be made without ensuring the testbenches are kept up to date.  Basic sanity means the RTL and TB are compatible and work together.  This must always be true.  So if a designer needs to run a sanity test (which will always be supported), they need the core-v-verif project.

davideschiavone:
I would keep the tb folder in there as many people may want to use the core with a minimal setup to run a `hello world` simple  C program without cloning other repositories.
Rather than removing such folder, I would simply change its name and meaning.
It's correct that this means maintaining that folder too, but it's so simple (and we can even make it simpler) that it's a good feature to have.

A name could be something that tells the user that in there you can run simple program. Any idea?

@bluewww , @MikeOpenHWGroup

bluewww:
I'm also for keeping the folder but the meaning should be downgraded from `testbench` to something like `example usage`.

davideschiavone:
I like `example`, but it doesn't explain a lot.
How about `sim` ?

MikeOpenHWGroup:
In core-v-verif we use "sim" to run tests.  The only committed file in "sim" is the Makefile - this keeps all simulation generated files in one place.

If you _must_ keep it, you could call it `example_tb`.   But I **_strongly_** recommend against it.

MikeOpenHWGroup:
Hey everyone, we should not let this issue get "stale".  I am already getting questions about the 'redundant" testbench in this repo, so we either need to get rid of it, or rename it to be clear that its not a production testbench.

davideschiavone:
I would still call it "sim" as it is actually its meaning. It runs tests/programs.
I propose otherwise `exe` or  `test_program`.  
Let's pick one and move forwards

MikeOpenHWGroup:
I'm not in favor of this.  The cv32e40p cannot be considered "sane" if it
doesn't work in its proper test bench.   So keeping a test bench in that
repo is a bad idea, regardless what we call it.

---mike

davideschiavone:
[...]

PR #238, opened 2020-02-19
renamed tb forlder to test_program #229

davideschiavone:
Renamed the tb folder to test_program


bluewww:
I mean if the name is a problem I would suggest to just call it 'example'. It is not a testbench anymore, just an example. Then you can also write in the README that this subdirectory is only for demonstration purposes.

My personal take on testing philosphy: It is a matter of the process. Mention multiple times in the README that people need to use core verif. Require people to add a test that covers the changes/feature. Require all CI to pass. Enforce code style and review. Demand big PR be split up etc. I dont think naming things a certain way is too important.

davideschiavone:
I do agree with @bluewww , it's not  a folder for `testing` or `verification`. 
It's just an example to let the users try out the core.
Also other cores have a directory where you can run Hello world. It does not mean your are testing it. We can call it example if it is better.

MikeOpenHWGroup:
> if the name is a problem I would suggest to just call it 'example'.

The name is not really the problem.  The concern is the existence of an unsanctioned testbench which discourages use of the real testbench by RTL Contributors.

> It is not a testbench anymore, just an example. Then you can also write in the README that this subdirectory is only for demonstration purposes.

The [core](https://github.com/openhwgroup/core-v-verif/tree/master/cv32/tb/core) testbench in [core-v-verif](https://github.com/openhwgroup/core-v-verif), and associated [README](https://github.com/openhwgroup/core-v-verif/blob/master/cv32/sim/core/README.md) in the `cv32/sim/core` directory should provide a simple example for Contributors to the CV32E40 RTL.

Silabs-ArjanB:
Hi @davideschiavone  Unfortunately this PR has been laying around for a very long (much too long) time and it is now completely outdated. Could you please either remove the pull request or issue a new one with changes that are still relevant?

davideschiavone:
Sure! Thanks for reminding it.

Issue #248, opened 2020-03-02
Rename clock gating module


davideschiavone:
The file is called *sim*, which is correct as it is meant to be a simulation-only file.
Not for synthesis.

The module though, should not have the sim name as it should be replace by a technology file where the real clock gating cell should be instantiated instead.

In that case, the tech file should be a SV file called cv32e40p_techname_clock_gating.sv which instantiate the CG cell. The module name of the wrapper should be the same as the simulation one.

PR #250, opened 2020-03-03
:art: Rename clock gating module #248

davideschiavone:
Fix for https://github.com/openhwgroup/cv32e40p/issues/248

### DIFF
--- a/rtl/cv32e40p_sim_clock_gating.sv
+++ b/rtl/cv32e40p_sim_clock_gating.sv
@@ -8,7 +8,7 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
-module cv32e40p_sim_clock_gating
+module cv32e40p_clock_gating
 (
     input  logic clk_i,
     input  logic en_i,
@@ -31,4 +31,4 @@ module cv32e40p_sim_clock_gating
   assign clk_o = clk_i & clk_en;
 `endif
 
-endmodule // cv32e40p_sim_clock_gating
+endmodule // cv32e40p_clock_gating
--- a/rtl/riscv_register_file_latch.sv
+++ b/rtl/riscv_register_file_latch.sv
@@ -114,7 +114,7 @@ module riscv_register_file
    // WRITE : SAMPLE INPUT DATA
    //---------------------------------------------------------------------------
 
-     cv32e40p_sim_clock_gating CG_WE_GLOBAL
+     cv32e40p_clock_gating CG_WE_GLOBAL
      (
       .clk_i     ( clk             ),
       .en_i      ( we_a_i | we_b_i ),
@@ -175,7 +175,7 @@ module riscv_register_file
    generate
       for(x = 1; x < NUM_TOT_WORDS; x++)
         begin : CG_CELL_WORD_ITER
-             cv32e40p_sim_clock_gating CG_Inst
+             cv32e40p_clock_gating CG_Inst
              (
               .clk_i     ( clk_int                               ),
               .en_i      ( waddr_onehot_a[x] | waddr_onehot_b[x] ),
--- a/rtl/riscv_register_file_latch.sv
+++ b/rtl/riscv_register_file_latch.sv
@@ -114,7 +114,7 @@ module riscv_register_file
    // WRITE : SAMPLE INPUT DATA
    //---------------------------------------------------------------------------
 
-     cv32e40p_clock_gating CG_WE_GLOBAL
+     cv32e40p_clock_gate CG_WE_GLOBAL
      (
       .clk_i     ( clk             ),
       .en_i      ( we_a_i | we_b_i ),
@@ -175,7 +175,7 @@ module riscv_register_file
    generate
       for(x = 1; x < NUM_TOT_WORDS; x++)
         begin : CG_CELL_WORD_ITER
-             cv32e40p_clock_gating CG_Inst
+             cv32e40p_clock_gate CG_Inst
              (
               .clk_i     ( clk_int                               ),
               .en_i      ( waddr_onehot_a[x] | waddr_onehot_b[x] ),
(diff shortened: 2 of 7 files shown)


rtl/riscv_register_file_latch.sv
```diff
@@ -114,7 +114,7 @@ module riscv_register_file
    // WRITE : SAMPLE INPUT DATA
    //---------------------------------------------------------------------------
 
-     cv32e40p_sim_clock_gating CG_WE_GLOBAL
+     cv32e40p_clock_gating CG_WE_GLOBAL
```
Silabs-ArjanB: Please do no use captitals in instance names

rtl/riscv_register_file_latch.sv
```diff
@@ -175,7 +175,7 @@ module riscv_register_file
    generate
       for(x = 1; x < NUM_TOT_WORDS; x++)
         begin : CG_CELL_WORD_ITER
-             cv32e40p_sim_clock_gating CG_Inst
+             cv32e40p_clock_gating CG_Inst
```
Silabs-ArjanB: Please do no use captitals in instance names

rtl/cv32e40p_sim_clock_gating.sv
```diff
@@ -8,7 +8,7 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
-module cv32e40p_sim_clock_gating
+module cv32e40p_clock_gating
```
Silabs-ArjanB: And filename has to match module name (so both need to become cv32e40p_clock_gate)

davideschiavone: I made them different as the file name, only in this case (exeptional case for non synthesizable modules), the file reminds the user that it's for simulation only.
When one synthesises the core in the technology node `tech123`, one creates a new file called `cv32e40p_tech123_clock_gating` that has inside a module called `cv32e40p_clock_gating` .
It instantiates the tech123 CG cell. That's why the difference between module and file name.
This is true only for non synthesizable (or actually, tech dependent only) files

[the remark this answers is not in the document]
davideschiavone: why that?
Ibex uses gating, https://github.com/lowRISC/ibex/blob/master/rtl/ibex_core.sv#L291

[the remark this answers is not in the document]
Silabs-ArjanB: Simply because it is a 'clock gate', not a 'clock gating'. You would call an instantiate and gate also simply 'and gate' and not 'and gating'.

Silabs-ArjanB: Okay, I am fine with non-equal file-module name then

Silabs-ArjanB: The end user indeed needs to modify the content of this file (e.g. by instantiating their technology dependent clock gate). In the end (after such modification) the file will still get synthesized, so I would prefer to stick to the equal filename - module name rule. If you want to prevent that people accidentally synthesize the original RTL, then we could consider something like ifndef SYNTH around the RTL and put a non-synthesizable 'replace me warning' in the other branch of the ifndef such that synthesis will break if this file is not updated by the user

davideschiavone: No actually this is not the idea. The idea is to not modify the core when synthesizing so that making pull requests or merging from the source is still easy. Simply, by creating a new file (that is local and never pushed), the simulation and synthesis script will simply compile and link the tech dependent module rather than the simulation only file. 
So I want to stress the fact that the file is for simulation only and cannot be synthesized (better redundant in this case as it's an important mistake one could do :) ) 
I agree about the warning message for synthesis instead. That's a great idea.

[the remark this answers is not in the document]
davideschiavone: Yes I agree. So let's differentiate that. Maybe @volgelpi wants to do the same for Ibex

rtl/cv32e40p_sim_clock_gating.sv
```diff
@@ -8,7 +8,7 @@
...
 (
     input  logic clk_i,
     input  logic en_i,
```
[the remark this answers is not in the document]
davideschiavone: We can solve this issue in another PR

rtl/riscv_register_file_latch.sv
davideschiavone: We can solve this issue in this thread https://github.com/openhwgroup/cv32e40p/issues/252

rtl/riscv_register_file_latch.sv
davideschiavone: We can solve this issue in this thread https://github.com/openhwgroup/cv32e40p/issues/252

rtl/cv32e40p_sim_clock_gating.sv
Silabs-ArjanB: Okay, now i understand. Sounds good to me. If we have a good manifest file for synthesis that will point to the 'to be created file by the end user' instead of the 'sim file', then we would not even need the ifndef synth. The user would immediately see that there is a missing file that they are supposed to add.

rtl/cv32e40p_sim_clock_gating.sv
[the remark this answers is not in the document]
Silabs-ArjanB: okay

rtl/riscv_register_file_latch.sv
Silabs-ArjanB: okay

rtl/riscv_register_file_latch.sv
Silabs-ArjanB: okay

rtl/cv32e40p_sim_clock_gating.sv
davideschiavone: Yes we are working on it in another PR :)

davideschiavone:
@Silabs-ArjanB I updates with your comments

Issue #255, opened 2020-03-10
Tracer does not log fence.i instruction


Imperas:
@davideschiavone 
As discussed earlier, when executing the fence.i instruction, this is not reported by the tracer.
This causes the step/compare to lose synchronization
could we please have a fix to this so that the step and compare is consistent
Thx
Lee

Issue #256, opened 2020-03-10
Retire signal not correct for CSRs


GTumbush:
Opening this issue based on call March 10, 2020. 

The tracer does not support printing CSR's to the tracer file.  I attempted to compare cv32e40p mstatus_q to mstatus predicted by the Imperas ISS and experienced inconsistent timing of mstatus_q update for the mret and ecall instructions. 

The mret instruction updates mstatus_q to 8, 1 clock cycle after the retire signal (see below).
![image](https://user-images.githubusercontent.com/54555114/76337501-104c5380-62bd-11ea-8577-207e7e2b92ec.png)

The ecall instruction updates mstatus_q to 6, 2 clock cycles before the retire signal (see below).
![image](https://user-images.githubusercontent.com/54555114/76337665-4c7fb400-62bd-11ea-8264-8f375e004c6b.png)

The software to produce the above is in the attached zip file.
[riscv_arithmetic_basic_test_0.zip](https://github.com/openhwgroup/cv32e40p/files/4313734/riscv_arithmetic_basic_test_0.zip)

GTumbush:
Adding the cycles signal to make the mret and ecall instructions easier to find:
![image](https://user-images.githubusercontent.com/54555114/76339217-66ba9180-62bf-11ea-8130-c6292615865b.png)

![image](https://user-images.githubusercontent.com/54555114/76339296-7fc34280-62bf-11ea-8739-f360a0d2df5d.png)

GTumbush:
wait(retire_delay) in riscv_tracer.sv does not compile with Incisive 15.20.077 or Xcelium 19.09.007

xmvlog(64): 19.09-s007: (c) Copyright 1995-2019 Cadence Design Systems, Inc.
      wait(retire_delay);
                      |
xmvlog: *E,EVNOTR (/wrk/gtumbush/mcu/riscy/cv32e40p_fix256/riscv/rtl/riscv_tracer.sv,854|22): An event is not a legal rvalue in this context [9.7.3(IEEE)].

Change to @(retire_delay);

GTumbush:
This fixes the issue with mret but not ecall.  

This is the new waveform for mret. Looks good!
![image](https://user-images.githubusercontent.com/54555114/76981667-888bc800-6900-11ea-9eca-19bb03344274.png)

Waveform for ecall is the same

GTumbush:
[...]

PR #261, opened 2020-03-18
:bug: tracer for fence instructions #255

davideschiavone:
Fixing issue "Tracer does not log fence.i instruction #255"

### DIFF
--- a/rtl/riscv_core.sv
+++ b/rtl/riscv_core.sv
@@ -1177,6 +1177,7 @@ module riscv_core
     .dret           ( id_stage_i.controller_i.dret_insn_i  ),
     .ecall          ( id_stage_i.controller_i.ecall_insn_i ),
     .ebreak         ( id_stage_i.controller_i.ebrk_insn_i  ),
+    .fence          ( id_stage_i.controller_i.fencei_insn_i),
     .rs1_value      ( id_stage_i.operand_a_fw_id           ),
     .rs2_value      ( id_stage_i.operand_b_fw_id           ),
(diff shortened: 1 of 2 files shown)


davideschiavone:
@imperas can you please double check if this solves your issue?

Imperas:
@GTumbush 
Greg, I think you should probably take first look at this

GTumbush:
What application do I run to test the fence instruction?

Imperas:
Ah, good question, this was the example from the google generator
I changed the hello.c example to add the following inline assembler
in the old tracer, this causes it to get out of sync

> int main () {
>     *asm("fence.i\n");*
>     printf("Hello World\n");
> }

That should be enough to show if the tracer is fixed

Thx
Lee

GTumbush:
I added asm("fence.i\n"); to hello.c and ran it against the fix255 code (cloned with git clone -b fix255 https://github.com/davideschiavone/riscv.git) and the hello test now passes with no miscompares.    The RTL treats fence as a no-op. The issue appears to be solved. Imperas, please close if you agree.

Imperas:
@davideschiavone @GTumbush 
Looks like this has solved the issue with the fence instructions not being traced - great.

Are you any further forward with the issues regarding CSR values at certain instructions and exceptions ?
Thx
Lee

GTumbush:
There have been commits but none have fixed the CSR issue reported in https://github.com/openhwgroup/cv32e40p/issues/256

davideschiavone:
I consider this issue closed then and we will focus on #256

PR #262, opened 2020-03-18
:bug: fix tracer retire event for xret instructions #256

davideschiavone:


davideschiavone:
@GTumbush , can you please give us feedback about the fix proposed?
it should solve the issue discussed during the phone called mentioned here
https://github.com/openhwgroup/cv32e40p/issues/256

GTumbush:
Hi Davide,
  No, sorry this did not fix the issue with the ecall instruction and created other issues.  

First regarding ecall. The addi instruction is now getting skipped for comparison because there is no retire event.  
image

The new issue is that there is an extra retire event at the beginning of execution. See below:
Original fix256 (commitish 9c8d876 )
image

Latest fix256 (commitish 0dccc0f)
image

It would help debugging if you do not force push. It took a lot of git trickery from John Martin to get the original fix256 code back.

davideschiavone:
Ok I won't do push --force for the next time! Sorry!

Can you please share the piece of code with addi and ecall? I can try to reproduce it and show the waveform I get.

Thanks
Davide

GTumbush:
You can use the riscv_arithmetic_basic_test_0.zip attached to the issue. https://github.com/openhwgroup/cv32e40p/issues/256

GTumbush:
Hi Davide,
 Were you able to use riscv_arithmetic_basic_test_0.zip attached to issue #256 to reproduce addi and ecall?
Thanks,
Greg

davideschiavone:
Dear @GTumbush , the addi is actually triggering one cycle before as I moved the retired event 
in the virtual ID/EX pipeline process instead of the EX/WB.
It actually not correct for load instructions by the way (my bad). 
So the problem is the following (given the intiial one)

```
Normal Instruction
ecall
```

The Normal instruction is in the ID stage, then EX stage, then the cycle later it triggers the RETIRE event, meanwhile the `ecall `is one cycle later, so when the `Normal Instruction` is at the EX stage, the ecall tells the CSRs to change values, the cycle later the ` Normal Instruction` triggers the retire event, but the ecall effects are already visible.

<table style="width:100%">
  <tr>
    <th>time</th>
    <th>ID stage</th>
    <th>EX stage</th> 
    <th>WB stage</th>
  </tr>
  <tr>
    <th>1</th>
    <th>Normal Instruction</th>
    <th>-</th>
    <th>-</th>
  </tr>
  <tr>
    <th>2</th>
    <th>ecall tells CSRs to change</th>
    <th>Normal Instruction</th>
    <th>-</th>
  </tr>
  <tr>
    <th>3</th>
    <th>Buble</th>
    <th>ecall, CSRs have changed</th>
    <th>Normal Instruction -> RETIRE</th>
  </tr>
</table>

The problem is at time 3. One solution is to post-pone in the RTL the command to tell the CSRs to change. I won't do it as this requires RTL changes that may have impact in the interrupts/security/debug. These things should be tested in depth to change such a sensitive this.
I am ok to do this but we have to keep it in a branch, so that we merge only once we have a full set of tests that cover the aforementioned features. Would this be a viable solution for you? If so I can make the change, and hopefully this will fix the problem

GTumbush:
Hi Davide,
I’m unclear on what you want to do. You say “I won't do it” and then “I am ok to do this”.  Are you OK to make the RTL change in a branch? If so, I’m OK with that.

Greg Tumbush

davideschiavone:
Yes sorry a bit confusing, I mean. I would like not to change the RTL for this fix, but I am ok to try 👍 

Please clone first in my fork of the core in the branch fix256

https://github.com/davideschiavone/riscv/tree/fix256

At the commit 6f90336bf6fe85e79b8019128f2f70d08a1dd676 you find a fix that should recreate the problem described up (reverting the last broken fix)

and in the commit cad0c7625fab468f5e3639cff8cd9be3b8862501 a further commit to change the RTL

Waiting for your feedback,
Davide

davideschiavone:
I would also like to know @bluewww 's opinion on this commit
cad0c7625fab468f5e3639cff8cd9be3b8862501

GTumbush:
Hi Davide,
   This looks much better. I executed the riscv_arithmetic_basic_test_0.zip attached to issue #256 and saw no miscompares.  

image

Greg

davideschiavone:
amazing! So I keep the PR open until @MikeOpenHWGroup will get a quite confident coverage to test the RTL changes I have done. If they pass, we will merge into master. How about that?

GTumbush:
Hi Davide, 
    OK, we can work off this branch for now. Shall I close issue 256 or wait for the code to be merged with the master?
Thanks,
Greg

davideschiavone:
Hello,
no let's keep it open as it's still an issue in the master branch, which is the reference.
Thanks
Davide

MikeOpenHWGroup:
> So I keep the PR open until @MikeOpenHWGroup will get a quite confident coverage to test the RTL changes I have done. If they pass, we will merge into master.
Thanks @davideschiavone.  I just ran a manual regression against the RTL at https://github.com/davideschiavone/riscv/tree/fix256.   All looks good!

bluewww:
@davideschiavone https://github.com/openhwgroup/cv32e40p/commit/cad0c7625fab468f5e3639cff8cd9be3b8862501 looks reasonable. CI in this repo doesn't seem to run. Did someone try to run the the `tb/dm` tests? Otherwise I give it a go.

davideschiavone:
No I did not tried, I think I have some problems running it currently on my ETH setup. Can you give it a try please @bluewww ?

Still, these modifications become critical when stalls/debug/interrupts show up mixing up

Silabs-ArjanB:
@davideschiavone Could you please add a drawing of how CSR instructions are supposed to be handled in the new/proposed RTL. This will help in reviewing the changes made to the controller (it is hard to review just based on these RTL changes by itself). We will need some kind of documentation that shows for all instruction groups when (i.e. in what pipeline stage) they retire and to indicated what is done to prevent that instructions retire out of order (or an indication on why that would be allowed).

davideschiavone:
I have mistakenly merged this PR and revert it. @Silabs-ArjanB , yes can you open a ticket for that?
Also, I think we should decide where to align all the CSR write/read operations in the pipeline and align the ones touched by the instructions as much as possible

Silabs-ArjanB:
Added https://github.com/openhwgroup/cv32e40p/issues/296 to document retirement policy for instructions.

Issue #253, opened 2020-03-06
Move CLINTx CSR addresses to custom space


davideschiavone:
Proposed addresses are 

MIEX = 0x7d0
MTVECX = 0x7d1
MIPX = 0x7d2

Silabs-ArjanB:
Please see https://github.com/openhwgroup/cv32e40p/pull/275

Silabs-ArjanB:
Now also updated documentation, please see https://github.com/openhwgroup/cv32e40p/pull/279

davideschiavone:
@vogelpi, the cv32e40p has been updated thanks to @Silabs-ArjanB 's effort! Thanks a lot @Silabs-ArjanB and thanks @vogelpi for pointing it out!

PR #275, opened 2020-03-27
Moved CLINTx CSR addresses to custom space #253

Silabs-ArjanB:

### DIFF
--- a/rtl/include/riscv_defines.sv
+++ b/rtl/include/riscv_defines.sv
@@ -207,15 +207,15 @@ typedef enum logic[11:0] {
   CSR_MISA      = 12'h301,
   CSR_MIE       = 12'h304,
   CSR_MTVEC     = 12'h305,
-  CSR_MIEX      = 12'h306,
-  CSR_MTVECX    = 12'h307,
+  CSR_MIEX      = 12'h7D0,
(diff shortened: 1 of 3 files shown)


davideschiavone:
Hi @Silabs-ArjanB , thanks a lot for the fix. Can you also update the the tb/core/ examples?

For instance, the crt0.s ues 0x306 as address
https://github.com/openhwgroup/cv32e40p/blob/master/tb/core/custom/crt0.S#L33

or the interrupt tests uses 307 for the MIPX.

Thanks a lot,
Davide

Silabs-ArjanB:
Of course, I didn't realize there was a test. Where do you normally run this test from (directly in the cv32e40p repos somehow or from core-v-verif)?

davideschiavone:
Ther are in the CV32e40p repo now but @MikeOpenHWGroup  is moving them to the core-verif for what concern the interrupt test. The custom folder only contains an hello world example and it will stay there as it is not a test. Indeed the tb directory will also change the name as it's not a "test" bench anymore.

Silabs-ArjanB:
Updated the assembly files now

vogelpi:
Thanks @davideschiavone for keeping me in the loop!

Issue #254, opened 2020-03-09
mtvec is tied to boot_addr_i when PULP_SECURE=0


jm4rtin:
When user mode is not added to the core (PULP_SECURE=0) the mtvec CSR has a fixed value (boot_addr_i) that can't be changed. We would like the ability to change this address even if user mode is not enabled in the core.

davideschiavone:
Dear all, @Silabs-ArjanB , @sjthales , @haugoug   what do you think about it?
Having the MTVEC read-only is ok by SPEC and it gives the possibility to save one 32b register.
I would like to ear your opinion to decide whether we want to have it READ-ONLY like today or writable like @jm4rtin  proposes.

bluewww:
I think for `PULP_SECURE=1` the `mtvec` register is 24 bits (lower 8 bits zeroed). In my opinion it makes sense to also make `mtvec` dynamic for `PULP_SECURE=0`.

There is obviously always the last possibility of making this behavior (hardcoded mtvec vs dynamic mtvec) configurable with a parameter, though I'm not sure how happy the dv team is about another parameter.

Silabs-ArjanB:
I would definitely prefer to make it possible that mtvec differs from boot_addr_i when PULP_SECURE=0. I would be fine with either adding a new input signal (mtvec_addr_i) or making mtvec(x) an actual register (same as PULP_SECURE=1. I have a preference for making this identical to the PULP_SECURE=1 behavior.

davideschiavone:
It makes sense to me that we do as PULP_SECURE=1. 
@jm4rtin would you like to prepare a PULL REQUEST? @Silabs-ArjanB and I will review it gladly

davideschiavone:
https://github.com/openhwgroup/cv32e40p/pull/285

Issue #277, opened 2020-03-27
Clean up parameters and pinout


Silabs-ArjanB:
Make non-supported parameters localparam and remove associated pins as agreed in https://github.com/openhwgroup/core-v-docs/blob/master/cores/cv32e40p/CV32E40P_and%20CV32E40_Features_Parameters.pdf

Silabs-ArjanB:
Two pull requests are associated with this:

https://github.com/openhwgroup/cv32e40p/pull/278
https://github.com/openhwgroup/core-v-verif/pull/85

Silabs-ArjanB:
Parameters/pins have been changed according to https://github.com/openhwgroup/core-v-docs/blob/master/cores/cv32e40p/CV32E40P_and%20CV32E40_Features_Parameters.pdf. Some new pin related chnages are being discussed, but they have their own github issue number.

PR #278, opened 2020-03-27
Parameter and pinout cleanup #277

Silabs-ArjanB:


Silabs-ArjanB:
Needs to be done at the same time as https://github.com/openhwgroup/core-v-verif/pull/85

Silabs-ArjanB:
Added back in the APU interface.

davideschiavone:
Thanks a lot @Silabs-ArjanB ,

few comments here:

1. The tb/dm testbench hasn't been updated with the new interface and parameters

2. Fixing PULP_SECURE to 0 makes the MTVEC(and MTVECx) tight to Boot Address, this makes the tests failing. It's a good example for issue #254
We need to fix #254 before having this PR merged. So I opened in parallel a PR for #254 , then you can merge it into your branch to have this point solved.

Silabs-ArjanB:
The status of the tb/dm is unclear to me. They already seemed out of date to begin with. Can you let me know which ones are actually in use and I will update them (some of them seem to have been copied into the core-v-verif repos instead).

Do you want me to work on #254?

davideschiavone:
PR for #254  is already here: https://github.com/openhwgroup/cv32e40p/pull/285
You just need to review.

tb/dm should just compile and work as before. The purpose is to provide an example to use GDB.
The `tb` folder will be replaced with the `example` name as its content is not for verification but just to provide quick examples. One with GDB/Debug and one with hello world.

davideschiavone:
@Silabs-ArjanB , as the PR #285 has been merged to master, the point 2. should be solved.

>     2. Fixing PULP_SECURE to 0 makes the MTVEC(and MTVECx) tight to Boot Address, this makes the tests failing. It's a good example for issue #254
>        We need to fix #254 before having this PR merged. So I opened in parallel a PR for #254 , then you can merge it into your branch to have this point solved.

You should now merge the master to your branch and see if it solves the issue and address the point 1 here:

>     1. The tb/dm testbench hasn't been updated with the new interface and parameters

Thanks @Silabs-ArjanB

Silabs-ArjanB:
Cleaned up parameters and pinout in various testbenches. 

- make vsim-run in 'cv32e40p/tb/dm' now compiles without errors and warnings. C code comilation however fails (I expect it was already like this)

- make custom-vsim-run in cv32e40p/tb/core now compiles without errors and warnings and executes okay as well.
- I don't have verilator so I couldn't check those compile targets (but updated all testbenches)
- make firmware-vsim-run-gui in cv32e40p/tb/core has gcc compilation issues (maybe a Makefile issue as 64-bit code is getting compiled it seems).

Could we open another ticket to get rid of the high amount of testbenches and duplication in cv32e40p. For example riscv_core is currently being instantiated in all of the following:

cv32e40p/tb/core/riscv_wrapper.sv
cv32e40p/tb/dm/tb_test_env.sv
cv32e40p/tb/dm/tb_top_verilator.sv
cv32e40p/tb/verilator-model/top.sv
cv32e40p/tb/tb_riscv/tb_riscv_core.sv

and riscv_wrapper for example is then being instantiated in 

tb/core/tb_top.sv 
tb/core/tb_top_verilator.sv

I updated all of the above but we should try to get rid of obsolete testbenches.

davideschiavone:
Hi @Silabs-ArjanB , my answers inline
> 
> Cleaned up parameters and pinout in various testbenches.
> 
>     * make vsim-run in 'cv32e40p/tb/dm' now compiles without errors and warnings. C code comilation however fails (I expect it was already like this
It works for me

> 
>     * make custom-vsim-run in cv32e40p/tb/core now compiles without errors and warnings and executes okay as well.

👍 

> 
>     * I don't have verilator so I couldn't check those compile targets (but updated all testbenches)
> 
>     * make firmware-vsim-run-gui in cv32e40p/tb/core has gcc compilation issues (maybe a Makefile issue as 64-bit code is getting compiled it seems).

The problem is the RVA instructions emitted. As you tight to 0 the RVA parameter, you should also disable those instructions.
To do so, please remove them from firmware/start.S and from the Makefile variables
`RISCV_TEST_INCLUDES ` and `FIRMWARE_TEST_OBJS `

> 
> 
> Could we open another ticket to get rid of the high amount of testbenches and duplication in cv32e40p. For example riscv_core is currently being instantiated in all of the following:
> 
> cv32e40p/tb/core/riscv_wrapper.sv
> cv32e40p/tb/dm/tb_test_env.sv
> cv32e40p/tb/dm/tb_top_verilator.sv
> cv32e40p/tb/verilator-model/top.sv
> cv32e40p/tb/tb_riscv/tb_riscv_core.sv
> 
> and riscv_wrapper for example is then being instantiated in
> 
> tb/core/tb_top.sv
> tb/core/tb_top_verilator.sv
> 
> I updated all of the above but we should try to get rid of obsolete testbenches.
Yes please go ahead. We already have a ticket to change name to the folder and clean it up. @MikeOpenHWGroup is porting the interrupt test. Then, we should only have an example of application and the debug connected.

Silabs-ArjanB:
Hi @davideschiavone , thank you for your feedback.

I will fix the RVA instruction issue as you described.

Can you be more specific about which parts I can remove please? For example, can I remove all of the following?

Files: 

cv32e40p/tb/dm/tb_top_verilator.sv
tb/core/tb_top_verilator.sv

Directories:

cv32e40p/tb/verilator-model
cv32e40p/tb/serDiv
cv32e40p/tb/tb_MPU
cv32e40p/tb/tb_riscv

davideschiavone:
Hi @Silabs-ArjanB ,

Can we have that discussion in another thread/issue ?

rtl/riscv_core.sv
```diff
@@ -75,11 +58,11 @@ module riscv_core
...
-  output logic                         instr_req_o,
-  input  logic                         instr_gnt_i,
-  input  logic                         instr_rvalid_i,
-  output logic                  [31:0] instr_addr_o,
-  input  logic [INSTR_RDATA_WIDTH-1:0] instr_rdata_i,
+  output logic        instr_req_o,
+  input  logic        instr_gnt_i,
+  input  logic        instr_rvalid_i,
+  output logic [31:0] instr_addr_o,
+  input  logic [31:0] instr_rdata_i,
```
MikeOpenHWGroup, line 65: Why are we removing parameters here?  Even if these widths must be fixed, there are a _lot_ of magic numbers in the code and it would be a good idea to slowly replace them all.  How about:
```
output logic                          instr_req_o,
input  logic                          instr_gnt_i,
input  logic                          instr_rvalid_i,
output logic [INSTR_ADDR_WIDTH -1:0]  instr_addr_o,
input  logic [INSTR_RDATA_WIDTH-1:0]  instr_rdata_i,
```

rtl/riscv_core.sv
```diff
@@ -91,8 +74,6 @@ module riscv_core
   output logic [31:0] data_wdata_o,
```
MikeOpenHWGroup, line 74: More magic numbers.  Consider replacing with parameter, even if these are intended to be fixed values for cv32e.

rtl/riscv_core.sv
```diff
@@ -476,7 +476,7 @@ module riscv_core
     .N_HWLP              ( N_HWLP            ),
     .RDATA_WIDTH         ( INSTR_RDATA_WIDTH ),
     .FPU                 ( FPU               ),
-    .DM_HaltAddress      ( DM_HaltAddress    )
+    .DM_HALTADDRESS      ( DM_HALTADDRESS    )
```
MikeOpenHWGroup, line 479: Nice.  All caps for parameters is a good coding style.

rtl/riscv_core.sv
```diff
@@ -561,7 +561,7 @@ module riscv_core
     .A_EXTENSION                  ( A_EXTENSION          ),
     .APU                          ( APU                  ),
     .FPU                          ( FPU                  ),
-    .Zfinx                        ( Zfinx                ),
+    .PULP_ZFINX                   ( PULP_ZFINX           ),
```
MikeOpenHWGroup, line 564: Dito.

Silabs-ArjanB:
> 
> 
> Hi @Silabs-ArjanB ,
> 
> Can we have that discussion in another thread/issue ?

Sure

rtl/riscv_core.sv
Silabs-ArjanB: Only instr_rdata_i had a parameter. There is no need to use a parameter for instr_addr_o. This will always be 32 bits wide. The use of INSTR_ADDR_WIDTH is debatable; I would not be against it but at the same time both cv32e40p and cv32e40 will have it set to 32.

rtl/riscv_core.sv
Silabs-ArjanB: It is a 32-bit core. Let's not go adding even more parameters

Silabs-ArjanB:
Fixed Makefile and firmware/start.S to no longer use non-supported instructions (i.e. atomics). Note that riscv_tests/rv32uc/rvc.S needed to be changed (as was done before) as it led to an internal compiler error.

davideschiavone:
For me it's ok to MERGE, just hitting @bluewww to ask him about the last change you have done in the RVC (he is the expert :) )

tb/core/firmware/start.S
```diff
@@ -317,17 +317,6 @@ start:
...
-	TEST(amoand_w)
-	TEST(amomaxu_w)
-	TEST(amomax_w)
-	TEST(amominu_w)
-	TEST(amomin_w)
-	TEST(amoor_w)
-	TEST(amoswap_w)
-	TEST(amoxor_w)
-	TEST(lrsc)
-
```
bluewww, line 330: Why is this removed? These test should work.

davideschiavone: The RVA parameter has been set to 0 by default, so they generate illegal instructions now

tb/core/mm_ram.sv
```diff
@@ -715,7 +716,7 @@ module mm_ram
     .req_core_i         ( instr_req_i            ),
     .req_mem_o          ( rnd_stall_instr_req    ),
 
-    .addr_core_i        ( 32'(instr_addr_i)      ),
```
bluewww, line 718: Removing this will generate a port mismatch warning

tb/core/riscv_tests/rv32uc/rvc.S
```diff
@@ -1,7 +1,156 @@
...
+  .option pop
+
+  TEST_PASSFAIL
+
+RVTEST_CODE_END
+
+  .data
+RVTEST_DATA_BEGIN
 
-#include "../rv64uc/rvc.S"
```
bluewww, line 7: Is this a copy paste of the `../rv64uc/rvc.S` file? Why would you do this for this file, but then you don't do it for the others?

There is a very good reason the structure of the test files is like this (i.e. with the `#include`s from `rv64uc`). They are a verbatim import from https://github.com/riscv/riscv-tests meaning that whenever there is an update to the upstream test repository, you can easily import the diff. Now that you change the structure you have to apply fixes everytime you do this. This is just an unnecessary increase in maintenance cost.

All in all its inconsistent and a bad idea. I would highly suggest to revert this change and apply the fix to `../rv64uc/rvc.S` which I guess is removing the `.option norelax`, a fix I mistakenly did because I compiled with `-march=rv32im` instead of `imc` I suspect.

tb/core/riscv_wrapper.sv
```diff
@@ -42,7 +43,7 @@ module riscv_wrapper
     logic [3:0]                   data_be;
     logic [31:0]                  data_rdata;
     logic [31:0]                  data_wdata;
-    logic [5:0]                   data_atop;
+    logic [5:0]                   data_atop = 'b0;
```
bluewww, line 46: indicate bitwidth

tb/core/Makefile
```diff
@@ -371,7 +369,7 @@ riscv_tests/rv32um/%.o: riscv_tests/rv32um/%.S riscv_tests/riscv_test.h \
 
 riscv_tests/rv32uc/%.o: riscv_tests/rv32uc/%.S riscv_tests/riscv_test.h \
 			riscv_tests/macros/scalar/test_macros.h
-	$(RISCV_EXE_PREFIX)gcc -c -march=rv32im -g -o $@ \
+	$(RISCV_EXE_PREFIX)gcc -c -march=rv32imc -g -o $@ \
```
bluewww, line 372: good

tb/core/Makefile
```diff
@@ -83,15 +83,13 @@ RISCV_EXE_PREFIX         = $(RISCV)/bin/riscv32-unknown-elf-
 
 # firmware vars
 FIRMWARE                 = firmware/
-RISCV_TEST_INCLUDES      = -Iriscv_tests/ -Iriscv_tests/macros/scalar \
-				-Iriscv_tests/rv64ui -Iriscv_tests/rv64um -Iriscv_tests/rv64ua
```
bluewww, line 87: Revert this too, see comment below

tb/dm/Makefile
```diff
@@ -30,7 +30,7 @@ VLOG_FLAGS		= -pedanticerrors -suppress 2577 -suppress 2583
 VLOG_LOG                = vloggy
 
 VOPT			= vopt-$(VVERSION)
-VOPT_FLAGS		= -debugdb -fsmdebug -pedanticerrors #=mnprft
+VOPT_FLAGS		= -debugdb -fsmdebug +acc -pedanticerrors #=mnprft
```
bluewww, line 33: Ok this makes sense otherwise the defaults would be really useless.

tb/dm/tb_test_env.sv
```diff
@@ -138,35 +140,29 @@ module tb_test_env
...
         .apu_master_operands_o  (                       ),
         .apu_master_op_o        (                       ),
         .apu_master_type_o      (                       ),
         .apu_master_flags_o     (                       ),
-        .apu_master_valid_i     (                       ),
-        .apu_master_result_i    (                       ),
-        .apu_master_flags_i     (                       ),
-
+        .apu_master_valid_i     ( 1'b0                  ),
+        .apu_master_result_i    (  'b0                  ),
```
bluewww, line 149: indicate bitwidth. Should stay consistent in this repo.

tb/dm/tb_top_verilator.sv
```diff
@@ -136,15 +139,14 @@ module tb_top_verilator #(
...
         .apu_master_operands_o  (                       ),
         .apu_master_op_o        (                       ),
         .apu_master_type_o      (                       ),
         .apu_master_flags_o     (                       ),
-        .apu_master_valid_i     (                       ),
-        .apu_master_result_i    (                       ),
-        .apu_master_flags_i     (                       ),
-
+        .apu_master_valid_i     ( 1'b0                  ),
+        .apu_master_result_i    (  'b0                  ),
```
bluewww, line 148: indicate bitwidth

tb/tb_riscv/riscv_random_stall.sv
```diff
@@ -86,7 +87,7 @@ endclass : rand_data_cycles
 
 mailbox #(stall_mem_t) core_reqs          = new (4);
 mailbox #(stall_mem_t) core_resps         = new (4);
-mailbox core_resps_granted = new (4);
+mailbox #(logic)       core_resps_granted = new (4);
```
bluewww, line 90: Revert this if you don't want to break Xcelium. It looks like a bug but it actually fixes a bug. Maybe a comment is in order.

See b54d51a031f4a62a0e3c1a8aecb1ab469c2a652f

tb/verilator-model/top.sv
```diff
@@ -96,37 +97,30 @@ module top
...
       .apu_master_operands_o  (                       ),
       .apu_master_op_o        (                       ),
       .apu_master_type_o      (                       ),
       .apu_master_flags_o     (                       ),
-      .apu_master_valid_i     (                       ),
-      .apu_master_result_i    (                       ),
-      .apu_master_flags_i     (                       ),
-
+      .apu_master_valid_i     ( 1'b0                  ),
+      .apu_master_result_i    (  'b0                  ),
```
bluewww, line 106: indicate bitwidth

tb/core/mm_ram.sv
Silabs-ArjanB: The connections to riscv_random_stall were actually generating port mismatch warnings and they have been fixed by above change. The key is that I also added an ADDR_WIDTH parameter to riscv_random_stall so that it can be used in different context with potentially different address with requirements (so that we can make this warning free in every context).

tb/core/riscv_tests/rv32uc/rvc.S
Silabs-ArjanB: The ./rv64uc/rvc.S and ./rv32uc/rvc.S file are not the same. Even on the master branch this test case led to a gcc internal error, so the starting point was a repos in which test cases were not compiling correctly.

I looked at what this test looked like roughly half a year ago and I re-applied the same change to get it to work. (This might be related to this issue: https://github.com/riscv/riscv-gnu-toolchain/issues/445)

   diff ./rv64uc/rvc.S ./rv32uc/rvc.S
   10a11,12
   > #undef RVTEST_RV64U
   > #define RVTEST_RV64U RVTEST_RV32U
   16,18d17
   <   // https://github.com/riscv/riscv-gnu-toolchain/issues/445
   <   // mixed modes should turn of relaxation (?)
   <   .option norelax

If we keep the test as it was the entire test suite will not run.

tb/core/riscv_wrapper.sv
Silabs-ArjanB: Okay, will do

tb/core/Makefile
Silabs-ArjanB: Agreed this should be reverted (which has been done now). The rv64* includes are however not actually being used yet as I am still not using the originally intended rvc.S (see discussion above). Once we get the rvc.S issue solved, then at least RISCV_TEST_INCLUDES will be already have the correct value.

tb/dm/tb_test_env.sv
Silabs-ArjanB: Fixed (and made similar fixes in other places)

tb/dm/tb_top_verilator.sv
Silabs-ArjanB: Done

tb/verilator-model/top.sv
Silabs-ArjanB: done

davideschiavone:
Actually, these changes in the RISCV_TESTS suite should be done in the core-verif repository, the only purpose of this "testbench" should be make hello world pass. So it's not wasted work as can be ported to core-verif, but probably we can ignore the them here

tb/core/riscv_tests/rv32uc/rvc.S
bluewww: This might also be a `pulp-riscv-gnu-toolchain` bug. I believe that the test worked with `riscv-gnu-toolchain` but I would have to check again.

Silabs-ArjanB:
Only the assembly/makefile changes seem to have been merged. Many of the proposed RTL and test bench/case updates have not yet been merged to master. Will they be merged later?

For example riscv_core still has the SHARED_* parameters and the Zfinx and DM_HaltAddress parameters according to their old naming style.

Currently (after above merge to master) we are still having warnings when compiling the RTL code, e.g.:

# ** Warning: (vsim-3015) [PCDPC] - Port size (32) does not match connection size (22) for port 'addr_mem_o'. The port definition is at: ../../tb/tb_riscv/riscv_random_stall.sv(52).
#    Time: 0 ns  Iteration: 0  Instance: /tb_top/riscv_wrapper_i/ram_i/instr_random_stalls File: mm_ram.sv Line: 702
# ** Warning: (vsim-3015) [PCDPC] - Port size (32) does not match connection size (22) for port 'addr_mem_o'. The port definition is at: ../../tb/tb_riscv/riscv_random_stall.sv(52).
#    Time: 0 ns  Iteration: 0  Instance: /tb_top/riscv_wrapper_i/ram_i/data_random_stalls File: mm_ram.sv Line: 738

** Warning: tb_test_env.sv(113): (vopt-2685) [TFMPC] - Too few port connections for 'riscv_core_i'.  Expected 46, found 45.
** Warning: tb_test_env.sv(113): (vopt-2718) [TFMPC] - Missing connection for port 'data_atop_o'.

PR #287, opened 2020-03-31
Updated Debug and Trigger documentation in user manual

silabs-PaulZ:


jm4rtin:
As far as I know there is no debug triggers in the core. Is there a pending change to add those? If not, why are the trigger registers being added to the documentation?

silabs-PaulZ:
Hi John, 
We did briefly discuss this HW breakpoint feature in a meeting, but I should have first put it as a pending issue before making a pull request. So yes, this is a request to add the Trigger feature (the same implementation as seen on IBEX).

davideschiavone:
Hello @silabs-PaulZ , thanks a lot for the PR.
Can you also add the new CSR registers to the  `Table` 7: Control and Status Register Map` please?

Also, are these CSRs optional (controlled by a parameter like in Ibex) or always present?

Thanks
Davide

davideschiavone:
Thanks @silabs-PaulZ !!! Merged it!

PR #305, opened 2020-04-10
first version of Synopsys DC script

davideschiavone:


davideschiavone:
Hello @ludmila-bta , @MikeOpenHWGroup 

here is my first version of the DC script. What do you think about it?

Thanks a lot
Davide

ludmila-bta:
Hi @davideschiavone,
Looks good. Please see my 5 cents below:
1. Manifest and SDC (timing constraints) files should be used (as separate files) to make the scripts more readable. This is just the beginning and the scripts will get more complicated with time.
2. I would avoid renaming the design, so it would match the RTL.
3. The sequence should be something like:

- read the libraries
- read the manifest file
- elaborate 
- current_design (to go to the top level)
- link (to link the design)
- read the sdc file
- run the compile_ultra command 
- you may want to add also "optimize_netlist -area" to improve the area after compile_ultra

4. Also I would recommend adding "-no_seq_output_inversion" switch to the compile_ultra

PR #307, opened 2020-04-16
moved to different files find fist one and population count

davideschiavone:
I moved from the riscv_alu.sv the pop count and find first one modules into 2 different files.

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -34,6 +34,8 @@ sources:
     - rtl/riscv_apu_disp.sv
     - rtl/riscv_fetch_fifo.sv
     - rtl/riscv_L0_buffer.sv
+    - rtl/riscv_popcnt.sv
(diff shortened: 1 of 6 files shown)


MikeOpenHWGroup:
Once this PR is merged, the core-v-verif testbenches are pointing to a DUT that is three generations behind the head.  I'll update that later this week.

BTW, if you ever want to test your RTL changes against a new hash, branch or repo, here is an example:
```
$ cd <your_working_copy_of_core-v-verif>
$ cd cv32/sim/uvmt_cv32
$ make CV32E40P_REPO=https://github.com/davideschiavone/riscv CV32E40P_BRANCH=style_alu CV32E40P_HASH=head
```
The above assumes you have a shell environment variable `SIMULATOR` set to what you are using.  You can set it on the command line for make it you prefer.

PR #310, opened 2020-04-21
Propagated is_decoding signal to the APU_DISPATCHER to mask read and write dependency

OttG:
### FIX

Propagate the `is_decoding` signal from the `controller` to the `apu_dispatcher` to mask read and write dependency. There was a case that, when the `fetch_valid` signal in the `if_stage` was 0, the instruction in the `decode_stage` would not advance but that same instruction would go to the `ex_stage`. This would cause a false read dependency with the currently executed instruction and the instruction it self (since the instruction in the `id_stage` and `ex_stage` was the same). Masking it with the `is_decoding` signal fixes this performance bug avoiding an extra stall when this happened.

### DIFF
--- a/rtl/riscv_apu_disp.sv
+++ b/rtl/riscv_apu_disp.sv
@@ -45,6 +45,7 @@ module riscv_apu_disp (
   output logic                          stall_o,
 
   // dependency checks
+  input  logic                          is_decoding_i,
   input  logic [2:0][5:0]               read_regs_i,
   input  logic [2:0]                    read_regs_valid_i,
   output logic                          read_dep_o,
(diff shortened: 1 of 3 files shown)


davideschiavone:
It's good for me. @Silabs-ArjanB , @stmach any comments?

davideschiavone:
Hi Arjan,
**`is_deconding`** depends on **`instr_valid_i `** which is not the signal coming from the instruction bus but a flip flop from the IF/ID stage.

Silabs-ArjanB:
> 
> 
> Hi Arjan,
> **`is_deconding`** depends on **`instr_valid_i `** which is not the signal coming from the instruction bus but a flip flop from the IF/ID stage.

I missed that. If it comes from a flop, then I am fine with this change.

Issue #302, opened 2020-04-08
Retire signal not correct for misaligned accesses


GTumbush:
The attached code does a misaligned access to memory.  The address accessed is 0x47F31. According to the RI5CY User Manual:

*The LSU is able to perform misaligned accesses, meaning accesses that are not aligned on natural word boundaries. However, it needs to perform two separate word-aligned accesses internally. This means that at least two cycles are needed for misaligned loads and stores.*

However, riscv_tracer triggers the Retire event after the first word-aligned access. It should not trigger the Retire event until after the second word-aligned access.  In addition a nonsense value is written to the insn_regs_write queue.

Referring to the waveform below:
1. Prior to the lw x15,-40(x8) instruction being executed relevant GPR values are:
   1. x8= 0x47F80
   1. x15=0
1. Instruction lw x15,-40(x8) is flagged as Retired at time 1,683,552ns
   1. addr=0xF and value=0x47f31 is written to the insn_regs_write queue.
   1. This is correct.
1.  Instruction lw x15, 0(x15) is flagged as Retired at time 1,683,650ns. However,
    1. The 2nd word-aligned access has not completed
    1. A nonsense value of 0x47F has been written to the insn_regs_write queue
    1.  The expected value written to the insn_regs_write queue is 0x04030201
1. Instruction addi x11, x15, 0 is flagged as Retired at time 1,683,811ns
    1. addr=0xB and value=0x04030201 is written to the insn_regs_write queue.
    1. This is correct

![image](https://user-images.githubusercontent.com/54555114/78831364-be274c80-79a6-11ea-9dad-1fa253604aae.png)

Tracer output showing nonsense value written to x15
1683451           22041 00008468 fd842783 lw               x15, -40(x8)        x15=00047f31  x8:00047f80  PA:00047f58
             1683552           22043 0000846c 0007a783 lw               x15, 0(x15)         **x15=0000047f** x15:00047f31  PA:00047f31
             1683771           22046 00008470 00078593 addi             x11, x15, 0         x11=04030201 x15:04030201

Please assign to Davide

Use the following code:
[misalign.tar.gz](https://github.com/openhwgroup/cv32e40p/files/4452700/misalign.tar.gz)

eroom1966:
@davideschiavone 
Hi Davide, could you please provide some feedback on this issue that Greg and myself have been investigating
Thx
Lee

davideschiavone:
Working on it! Let you know asap!

[...]

PR #313, opened 2020-04-22
Fix#302

davideschiavone:
This PR addes fixes for retire event of misaligned memory accesses and xret and ecall functions

Issues: 

https://github.com/openhwgroup/cv32e40p/issues/302
https://github.com/openhwgroup/cv32e40p/issues/256


davideschiavone:
Ok so now the master should already have the fence trace and this PR should solve #302 and #256

davideschiavone:
@GTumbush confirmed that it works now.
Waiting for @Silabs-ArjanB to merge.

best
Davide

PR #315, opened 2020-04-23
riscv_cs_register: Fix clearing of PC registers

accuminium:
According to the RISC-V specification (Section 9.1 of the Unprivileged ISA), the CSR Read and Clear instruction (CSRRC) treats the value in register `rs1` as a bit mask that specifies bit positions to be cleared in the CSR.  Any bit that is high in `rs1` will cause the corresponding bit to be cleared in the CSR.

This commit changes the code for `PCCR`, `PCMR`, and `PCER` accordingly.


davideschiavone:
Thanks Andreas, pinging @Silabs-ArjanB for comments before merging

Silabs-ArjanB:
Changes look good; thanks!

Issue #318, opened 2020-04-29
Data address is not consistent with byte enable signals


Silabs-ArjanB:
In the second phase of a non-aligned load/store the data address (data_addr_o) is not consistent with the byte enables (data_be_o). 

For example when doing a word write to address 0xffff_f801 we will get two transfers:

- First transfer of 3 bytes: Address 0xffff_f801, byte lanes 4'b1110
- Second transfer of 1 byte: Address 0xffff_f805, byte lanes 4'b0001

![image](https://user-images.githubusercontent.com/40633348/80608737-835d8500-8a37-11ea-8563-0fd5cfe09a0a.png)

Slaves can of course simply ignore the 2 LSBs of the address. The OBI spec however requires the address to be consistent with the byte lanes (which will enable an easier translation to AXI). As the second transfer will always have its LSB set (data_be_o[0] is always 1), the 2 LSBs of the address must always be 0 for that second transfer.

The fix is easy (and I already did it on the pull request that I soon plan to issue):

Replace the following line in riscv_load_store_unit-sv:

  assign data_addr_o      = data_addr_int;

with:

  assign data_addr_o = data_misaligned_ex_i ? {data_addr_int[31:2], 2'b00} : data_addr_int;

Issue #257, opened 2020-03-12
Path from data_rvalid_i to instr_req_o on 128b prefetcher


davideschiavone:
There is a path in the prefetcher 128b wide that goes from the data_rvalid_i to the instr_req_o.

Such path limits signifantly the frequency in the PULP cluster.

Silabs-ArjanB:
Sounds similar to (or same as) https://github.com/openhwgroup/cv32e40p/issues/126

davideschiavone:
Hi @Silabs-ArjanB , not it's actually from **data**_valid to **instr**_req. So it's different.
Anyway, this is a specific PULP issue as the 128b won't be used by OHW for the moment

Silabs-ArjanB:
Hi @davideschiavone Can you please close this issue as the 128b prefetcher has been removed?

Issue #176, opened 2019-10-02
Problem on second requisition on misaligned stores


aignacio:
Hi everyone,
I have been designing a dbus to ahb converter for RI5CY cpu and I have a question related to set the converter to operate with two-cycle misaligned requisitions. So, considering that a memory or cache interface could handle misaligned requisitions through ahb bus, how did you make the conversion in terms of AHB Address considering the two LSBs  from the req/grant interface? I mean, as an example here below, have a word to store shifted by one byte into (0xb000_0000), so the first req. makes sense ahb addr == addr_i but in the second requisition to store the remaining byte, the addr_i it's equal to 'h5, what indeed would shift the data by one byte once you make ahb_addr == addr_i, IMO the addr should be 'h4 (nxt word boundary), isn't it? 
```
    li t0, 0xDEADBEEF
    li t1, 0xB0000000  
    nop
    sw t0, 0(t1)        # 0xB0000000 = DEADBEEF /  0xB0000004 = 00000000
    nop
    sw t0, 1(t1)        # 0xB0000000 = ADBEEFEF  /  0xB0000004 = 000000DE
```
![Screenshot from 2019-10-02 13-18-51](https://user-images.githubusercontent.com/5991372/66042954-2986f200-e51e-11e9-9d74-c265b7524b87.png)

Silabs-ArjanB:
Hi @aignacio, I agree with your observation and I had actually filed as similar ticket (#318) to ease conversion to protocols as AHB and AXI. If this pull request is accepted (https://github.com/openhwgroup/cv32e40p/pull/322) then the issue will have been fixed as you suggested.

Silabs-ArjanB:
Hi @aignacio This issue has been fixed now (with above pull request). If you agree can you please close this issue?

aignacio:
yeap, done

Silabs-ArjanB:
Thanks

Issue #129, opened 2019-08-26
Non-reachable FSM states in load/store unit


Silabs-ArjanB:
The WAIT_RVALID_EX_STALL and   IDLE_EX_STALL states in the load/store unit FSM are not reachable.

**Git tag:** 

eae1cabcdb2333e5f1037996a48504985e7813a7 (august 5   2019)

**Parameters:**

(PULP_SECURE, USE_PMP, PULP_CLUSTER)   = (0, 0, 0) or (1, 1, 0) or (0, 0, 1) or (1, 1, 1) 
N_EXT_PERF_COUNTERS = 0   
INSTR_RDATA_WIDTH = 32   
N_PMP_ENTRIES = 16   
FPU = 0   
Zfinx = 0   
FP_DIVSQRT = 0   
SHARED_FP = 0   
SHARED_DSP_MULT = 0  
SHARED_INT_DIV = 0   
SHARED_FP_DIVSQRT= 0  
WAPUTYPE = 0   
APU_NARGS_CPU = 3   
APU_WOP_CPU = 6   
APU_NDSFLAGS_CPU   = 15   
APU_NUSFLAGS_CPU   = 5   
DM_HaltAddress = 32'h08000000

The WAIT_RVALID_EX_STALL and IDLE_EX_STALL states of CS (current state) are not reachable for the above listed parameter configurations. Is this dead/obsolete code in the RTL, or are these states reachable for other parameter configurations, or are these states intended for future extensions to RI5CY?

davideschiavone:
We also noticed it at : https://github.com/pulp-platform/riscv/issues/80

We are investigating

davideschiavone:
removed in new_pipeline branch. Under evaluation

3bfe1ea820e60e3239a3d9b3c8ef403c45d4c3c1

PR #322, opened 2020-05-01
Arjan b obi fix base

Silabs-ArjanB:
This pull request makes the prefetch unit and the load/store unit OBI compatible. The external bus interface pins remained unchanged, but internally certain unwanted combinatorial paths have been removed.

Fixes are provided for the following issues: #126, #127, #128, #129, #318, #80, #50, #176   . In order to provide backward compatibility to the ETH Zurich PULP team, the fixes for #127 and #128 can be undone via the newly introduced localparam PULP_OBI. As part of this pull request also the 128-bit wide fetch interface has been removed (as agreed in https://github.com/openhwgroup/core-v-docs/blob/master/cores/cv32e40p/CV32E40P_and%20CV32E40_Features_Parameters.pdf), therefore also #17, #125, and #257 can be closed when this pull request is accepted.   

Local verification was done with the following tests:

- firmware-vsim-run (i.e. firmware test cases, RISC-V compliance tests)
- custom-vsim-run (with and without random wait states)
- interrupt-vsim-run (interrupt test case)
- locally modified (not-pushed) version of firmware-vsim-run with added random wait state support and locally fixed/enhanced versions of I-MISALIGN_LDST-01.S, I-SB-01.S, I-SH-01.S

Unwanted combinatorial paths have been proven removed by running synthesis; OBI compliance has been checked with formal verification (assertions for this will be contributed later), so achievable system frequency will be improved. Cycle count remained identical for firmware-vsim-run and custom-vsim-run when run without wait states. Cycle count improved/reduced by 8% and 14% for firmware-vsim-run and custom-vsim-run respectively when run with random wait states (max 10 wait states on both gnt and rvalid). (This cycle count improvement is okay of course, but not for a realistic scenario (as the number of wait states is unrealistically high). In principle the fixes can theoretically have a negative impact on branch performance as explained in #128;  reason for the overall reduced cycle count is that waits for gnt and waits for rvalid (for different transactions) can now occur in parallel (instead of sequential as in the original design).

This pull request is part of a multi-phase transformation as explained in the OpenHW TWG Cores channel. Temporarily hardware loops have been removed and these will be re-introduced in a following phase. (There is a new top level parameter called PULP_HWLP which for now has to be set to 0 and for which value 1 will re-enabled in a following phase. The actual instruction decoder still needs to take PULP_HWLP into account; there will be a separate pull request for that modification).

There will be a separate pull request related to documentation updates as well.


silabs-PaulZ:
As part of the multiphase pull request, I've added #327 pull request as a follow up to parameterized the Hardware Loop logic in the riscv_decoder.sv.

Silabs-ArjanB:
To aid in review the instruction prefetch buffer is now organized as follows:

image

MikeOpenHWGroup:
@Silabs-ArjanB said:
> There will be a separate pull request related to documentation updates as well.

Nice!   I took the liberty of adding the figure above to the user manual [images](https://github.com/openhwgroup/core-v-docs/tree/master/cores/cv32e40p/user_manual/images) directory in core-v-docs.  The [intro](https://github.com/openhwgroup/core-v-docs/edit/master/cores/cv32e40p/user_manual/source/intro.rst) chapter has an example of how to include these in restructured text.

Silabs-ArjanB:
No, the Genus synthesis tool actually craps out on $fatal. The assertions it does not like either, but there it will continue and generate warnings. In a future pull request I intend to clean up the used if(n)defs a little bit (for example I think all assertions should be directly guarded by ifndef SYNTHESIS and it would be good to have one specific define to enable or disable assertions).

MikeOpenHWGroup:
Learn something new every day.

I would recommend that the assertions all be removed from the RTL source and placed in a separate module/file that is bound in to the RTL.

PR #324, opened 2020-05-01
Cleaned up undriven signals and inferred latch (trap_base_addr)

Silabs-ArjanB:

Added 0 assignments to many signals that became undriven due to our chosen configuration. Also prevented latch generation on trap_base_addr (in IF stage) by adding default 0 assignment.

### DIFF
--- a/rtl/riscv_core.sv
+++ b/rtl/riscv_core.sv
@@ -128,8 +128,8 @@ module riscv_core
   // these used to be former inputs/outputs of RI5CY
 
   logic [5:0]                     data_atop_o;  // atomic operation, only active if parameter `A_EXTENSION != 0`
-  logic [N_EXT_PERF_COUNTERS-1:0] ext_perf_counters_i = 'b0;
-  logic                           irq_sec_i = 1'b0;
+  logic [N_EXT_PERF_COUNTERS-1:0] ext_perf_counters_i;
+  logic                           irq_sec_i;
   logic                           sec_lvl_o;
 
   localparam N_HWLP      = 2;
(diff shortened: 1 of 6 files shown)


davideschiavone:
Hello @Silabs-ArjanB , any comment on this?

Silabs-ArjanB:
Hi @davideschiavone , I am not sure what type of comment from me you are looking for. I made the pull request and I do not see any further questions.

davideschiavone:
I commented line 134 of this PR with the following:

> 
> 
> 
> Hello @Silabs-ArjanB , thanks for your PR. Shall we maybe remove the exc_pc and trap_base_addr at line 133 and 134 and add them in the default statement?
> We should also probably not assign them to 0 as it may add another level of complexity. I would assign them to { trap_base_addr, 8'h0 } and m_trap_base_addr_i respectively.
> 
> What do you think?
>

Silabs-ArjanB:
Hi @davideschiavone , your proposal is better / cheaper indeed; I will do a new commit

Silabs-ArjanB:
Hi @davideschiavone I did the fix you asked for. Weirdly enough I cannot see anywhere that you left a review comment. One warning here: Hello World works, but the 'firmware test' is broken because of the usage of old style CSRs related to performance counters inside stats.c. I am assuming that this is already fixed in some other pull request.

davideschiavone:
> 
> 
> Hi @davideschiavone I did the fix you asked for. Weirdly enough I cannot see anywhere that you left a review comment. One warning here: Hello World works, but the 'firmware test' is broken because of the usage of old style CSRs related to performance counters inside stats.c. I am assuming that this is already fixed in some other pull request.

We do not want to maintain them here. They will be removed. Here we maintain only hello world

Issue #130, opened 2019-08-26
CSR access to non-existing CSR


Silabs-ArjanB:
A CSR access to a non-existing CSR does not raise an illegal instruction exception. Similarly a write attempt to a read-only CSR does not raise an illegal instruction exception.

For more details see the attached pdf.

[RI5CY_CSR.pdf](https://github.com/pulp-platform/riscv/files/3541815/RI5CY_CSR.pdf)

Silabs-ArjanB:
Fixed

PR #326, opened 2020-05-03
Issue #130 : Generate illegal instruction when accessing CSR that is …

silabs-PaulZ:
Generate illegal instruction when accessing CSR that is not implemented or blocked (e.g. debug CSR). Clean up CSR defines.

### DIFF
--- a/rtl/riscv_decoder.sv
+++ b/rtl/riscv_decoder.sv
@@ -31,10 +31,12 @@ import riscv_defines::*;
 
 module riscv_decoder
 #(
+  parameter PULP_HWLP         = 0,
   parameter A_EXTENSION       = 0,
   parameter FPU               = 0,
   parameter FP_DIVSQRT        = 0,
   parameter PULP_SECURE       = 0,
+  parameter USE_PMP           = 0,
   parameter SHARED_FP         = 0,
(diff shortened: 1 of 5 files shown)


davideschiavone:
Looks ok for me. @Silabs-ArjanB ? Can we merge?

rtl/riscv_decoder.sv
```diff
@@ -2401,9 +2405,82 @@ module riscv_decoder
...
+            // User register access
+            CSR_USTATUS,
+              CSR_UTVEC,
+              CSR_UEPC,
+              CSR_UCAUSE :
+                if(!PULP_SECURE) csr_illegal = 1'b1;
+
+            default : csr_illegal = 1'b1;
+
+          endcase // case (instr_rdata_i[29:28])
```
Silabs-ArjanB, line 2471: instr_rdata_i[31:20])

davideschiavone:
OK thanks @Silabs-ArjanB , let's wait the fix of the typo from @silabs-PaulZ and then we merge

rtl/riscv_decoder.sv
silabs-PaulZ: Fix implemented

Issue #330, opened 2020-05-08
Add Debug Trigger Module Feature


silabs-PaulZ:
The Debug Trigger Module is a new feature, approved by the core's task group, and documented in the CSR and Debug CV32E40P documentation.

This issue to track the assignment of adding this new feature.

PR #331, opened 2020-05-11
#330 add debug trigger

silabs-PaulZ:
Add debug trigger logic resolving issue #330.

Trigger CSRs added:
  CSR_TSELECT   = 12'h7A0,
  CSR_TDATA1    = 12'h7A1,
  CSR_TDATA2    = 12'h7A2,
  CSR_TDATA3    = 12'h7A3,
  CSR_MCONTEXT  = 12'h7A8,
  CSR_SCONTEXT  = 12'h7AA,
Only TDATA1 and TDATA2 have RW fields, the rest are default 0

Trigger CSRs not accessible unless in debug_mode.

Added DEBUG_TRIGGER_EN parameter.

### DIFF
--- a/rtl/include/riscv_defines.sv
+++ b/rtl/include/riscv_defines.sv
@@ -256,6 +256,14 @@ typedef enum logic[11:0] {
   CSR_PMPADDR14 = 12'h3BE,
   CSR_PMPADDR15 = 12'h3BF,
 
+  // Trigger
+  CSR_TSELECT   = 12'h7A0,
+  CSR_TDATA1    = 12'h7A1,
+  CSR_TDATA2    = 12'h7A2,
+  CSR_TDATA3    = 12'h7A3,
+  CSR_MCONTEXT  = 12'h7A8,
+  CSR_SCONTEXT  = 12'h7AA,
+
   // Debug/trace
   CSR_DCSR      = 12'h7b0,
   CSR_DPC       = 12'h7b1,
@@ -449,6 +457,7 @@ parameter TRAP_USER         = 2'b01;
 parameter TRAP_MACHINEX     = 2'b10;
 
 // Debug Cause
+parameter DBG_CAUSE_NONE       = 3'h0;
 parameter DBG_CAUSE_EBREAK     = 3'h1;
 parameter DBG_CAUSE_TRIGGER    = 3'h2;
 parameter DBG_CAUSE_HALTREQ    = 3'h3;
@@ -466,6 +475,13 @@ parameter DBG_SETS_EBRK   = 1;
 parameter DBG_SETS_SSTE   = 0;
 
 parameter DBG_CAUSE_HALT   = 6'h1F;
+
+// Constants for the dcsr.xdebugver fields
+typedef enum logic[3:0] {
+   XDEBUGVER_NO     = 4'd0, // no external debug support
+   XDEBUGVER_STD    = 4'd4, // external debug according to RISC-V debug spec
+   XDEBUGVER_NONSTD = 4'd15 // debug not conforming to RISC-V debug spec
+} x_debug_ver_e;
 
 
 /////////////////////////////////////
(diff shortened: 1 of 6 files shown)


rtl/riscv_decoder.sv
```diff
@@ -2445,6 +2446,16 @@ module riscv_decoder
...
                 if(!debug_mode_i) csr_illegal = 1'b1;
 
+            // Debug Trigger register access
+            CSR_TSELECT     ,
+              CSR_TDATA1    ,
+              CSR_TDATA2    ,
+              CSR_TDATA3    ,
+              CSR_MCONTEXT  ,
+              CSR_SCONTEXT  :
+                if(!debug_mode_i || DEBUG_TRIGGER_EN != 1)
```
jm4rtin, line 2456: From section "5.2 Trigger Registers" of the debug spec:
> The trigger registers are only accessible in machine and Debug Mode to prevent untrusted user code from causing entry into Debug Mode without the OS's permission.

This makes them only accessible from debug mode.

silabs-PaulZ: Correct, the DMODE field is tied to a value of 1, only allow debugger access.

rtl/riscv_core.sv
```diff
@@ -40,6 +40,7 @@ module riscv_core
   parameter PULP_CLUSTER        =  0,
   parameter FPU                 =  0,
   parameter PULP_ZFINX          =  0,
+  parameter DEBUG_TRIGGER_EN    =  1,
```
Silabs-ArjanB, line 43: Why not unconditionally include this feature? We did not specify this to become a top level parameter.

rtl/include/riscv_defines.sv
```diff
@@ -256,6 +256,14 @@ typedef enum logic[11:0] {
   CSR_PMPADDR14 = 12'h3BE,
   CSR_PMPADDR15 = 12'h3BF,
 
+  // Trigger
+  CSR_TSELECT   = 12'h7A0,
+  CSR_TDATA1    = 12'h7A1,
+  CSR_TDATA2    = 12'h7A2,
+  CSR_TDATA3    = 12'h7A3,
+  CSR_MCONTEXT  = 12'h7A8,
+  CSR_SCONTEXT  = 12'h7AA,
```
Silabs-ArjanB, line 265: It is not fully clear to me from the Debug Spec which CSRs you are actually suppsoed to implement. My interpretation of the spec is that 'most trigger functionality is optional' but that does not mean (in my interpretation) that the CSRs should not be there (and lead to an illegal instruction exception). Can you comment on why TINFO, TCONTROL are not used here? And (slightly contradicting myself now), why do we have SCONTEXT, isn't that supposed to be present only in cores supporting S mode?

rtl/riscv_core.sv
silabs-PaulZ: This parameter was documented in the debug section and it matches the IBEX implementation. 
I agree that we should remove this parameter and always include Trigger since it does not take significant area to implement. I will make an update to this branch as well as change the documentation.

davideschiavone: I would make it local as @Silabs-ArjanB suggested, so that in case someone wants to fork the core and remove such feature has an easier modification to do

rtl/include/riscv_defines.sv
silabs-PaulZ: The Debug Spec mentions the following:

- `TCONTROL `is optional and applies to M-mode trap handler.
image

- `TINFO`: If `type `is not writable, this register may be unimplemented, in which case reading it causes an illegal instruction exception.
image

The other Trigger CSRs (such as SCONTEXT) are not listed as being optional, and therefore these simply return a value of 0.

Silabs-ArjanB: Thanks for the further explanation; I am good with the pull request then.

rtl/riscv_core.sv
silabs-PaulZ: localparam change has been updated to this branch

PR #332, opened 2020-05-12
tb: Remove inline ecall

bluewww:
These actually serve no purpose. They were used for local testing and
accidentally commited and propagated.

### DIFF
--- a/tb/core/custom/hello_world.c
+++ b/tb/core/custom/hello_world.c
@@ -44,8 +44,6 @@ int main(int argc, char *argv[])
 #ifdef RANDOM_MEM_STALL
     activate_random_stall();
 #endif
-    /* inline assembly */
-    asm volatile("ecall");
     /* write something to stdout */
     printf("hello world!\n");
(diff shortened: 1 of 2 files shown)


davideschiavone:
It looks ok thanks @bluewww ! @MikeOpenHWGroup ? any comment?

davideschiavone:
it's actually broken as the ecall is not needed and in general it is not saving the X1 register, so removing the printf would break it.
I showed you the PR as you may have the core-verif based on this test so I wanted to make you aware of this change. I would merge.

MikeOpenHWGroup:
> I showed you the PR as you may have the core-verif based on this test so I wanted to make you aware of this change. I would merge.

We have a lot of other issues like this in core-v-verif.  :wink: 

Anyway, I will approve this PR, but you should do the merge - its never a good idea to have verification people merging into the design code.

davideschiavone:
thanks @MikeOpenHWGroup :) done

Issue #333, opened 2020-05-12
Replace performance counters with Hardware Performance Monitor


silabs-PaulZ:
The following change was approved by the OpenHW Core Task Group:

Replace existing performance counter implementation with the Hardware Performance Monitor specified in the privileged architecture.

Access to an existing performance counter will be removed:
- PCMRs, 
- PCERs,
- PCCRs
No longer support external events

Specific feature details that were not fully discussed are now mentioned in this issue :

- All counters shall be 64bits.
- MCYCLE and MINSTRET shall always be implemented and available
- The number of MHPMCOUNTERs (and corresponding MHPMEVENTs) registers shall be parameterizable with a top level parameter, NUM_MHPMCOUNTERS
   - default value of 1
   - range between 1 and 29
   - Access to non-implemented MHPMCOUNTER and MHPEVENT shall be allowed and will always return a read value 0
- MCOUNTEREN and SCOUNTEREN shall not be implemented, as these WARL registers control access to lower privilege modes and shall return a value of 0 when read (it will not cause an illegal instruction)
- MCOUNTINHIBIT shall be implemented to disable/enable counting of each counter
- The MHPMEVENT will each have 15 bits, allowing for the selection of the following monitor counting options (matches existing PCCR count types):

Index | Name | Description
---|---|---
0 | CYCLES | Counts the number of cycles the core was active (not sleeping)
1 | INSTR | Counts the number of instructions executed
2 | LD_STALL | Number of load data hazards
3 | JR_STALL | Number of jump register data   hazards
4 | IMISS | Cycles waiting for instruction   fetches, i.e. number of instructions wasted due to non-ideal caching
5 | LD | Number of data memory loads   executed. (Misaligned accesses are counted twice)
6 | ST | Number of data memory stores   executed. (Misaligned accesses are counted twice)
7 | JUMP | Number of unconditional jumps   (j, jal, jr, jalr)
8 | BRANCH | Number of branches. (Counts taken and not taken branches)
9 | BTAKEN | Number of taken branches.
10 | RVC | Number of compressed   instructions executed
11 | FP_TYPE | Cycles wasted due to different   latencies of subsequent FP-operations
12 | FP_CONT | Cycles wasted due to   contentions at the shared FPU (PULP only)
13 | FP_DEP | Cycles wasted due to data   hazards in subsequent FP instructions
14 | FP_WB | Cycles wasted due to FP   operations resulting in write-back contentions

[...]

Issue #47, opened 2018-08-21
Core performance counter is not accounting all cycles


haugoug:
When using the CYCLES event, not all cycles are accounted while all cycles should be accounted as soon as the core is clocked.
It seems to be only with special instructions like csr instructions.
The issue can be reproduced with this test: tests/pulp_tests/bugs/perf_cycle_lost
This happens on all chips even vega.

haugoug:
I also attach the instruction trace which shows the issue:
      11435920000000           69900 1c008ba4 780027f3 csrrs            x15, x0, 0x780      x15=00007d5b
      11435940000000           69901 1c008ba8 40f707b3 sub              x15, x14, x15       x15=0000001d x14:00007d78 x15:00007d5b
      11435960000000           69902 1c008bac fef04ce3 blt              x0, x15, -8         x15:0000001d
      11436019000000           69905 1c008ba4 780027f3 csrrs            x15, x0, 0x780      x15=00007d5f
      11436039000000           69906 1c008ba8 40f707b3 sub              x15, x14, x15       x15=00000019 x14:00007d78 x15:00007d5f
      11436058000000           69907 1c008bac fef04ce3 blt              x0, x15, -8         x15:00000019

As you can see the iteration takes 5 cycles but the core counter is increased by only 4 at each iteration

silabs-PaulZ:
The performance counters will be changed into HPM, as defined by the privileged specification. This bug was also still seen in this conversion.

The bug came about from a CSR read (CSRR) being assembled into a CSR read & set instruction where the data is a value of 0 (CSRRS w/rs=0 or CSRRSI w/zimm=0). This would, in turn, cause a write-back operation, essentially preventing an increment of the counter when reading. 

The fix has been implemented in the pull request #339.

Silabs-ArjanB:
Hi @haugoug Could you please close this ticket if you agree it has been fixed?

PR #339, opened 2020-05-14
Fix for issue #333: Replace performance counters with HPM

silabs-PaulZ:
This pull request fixes issue #333: Replace the performance counters with Hardware Performance Monitor (HPM). 

There may be a separate pull request to add (back) the external events.

The sanity and hello_world test passes, but the firmware tests will fail (at a different location than normal) as it uses the old performance counters.


silabs-PaulZ:
The firmware tests fix can be found here:
https://github.com/openhwgroup/core-v-verif/pull/108

silabs-PaulZ:
This pull request also fixes #47.

PR #340, opened 2020-05-14
extend internal debug request until debug_mode is enterred

### LINKED ISSUE #233: RI5CY: User Manual v4.2-14.5.1 Vectorial ALU Operations table missing information
The Description for the Mnemonics is missing information to understand how the instructions operate. 
For example, the indexing variable "i" range and meaning are not defined; pv.*.sc.* operand use is not clearly described; carry/borrow handling are not described; and the use of the "0xFFFF" mask is undefined in all cases.
For example:
* pv.add.h is understood as:
  * rD[31-16] = rs1[31-16] + rs2[31-16] 
  * rD[15-0] = rs1[15-0] + rs2[15-0]
* Is pv.add.sc.h understood as?
  * rD[31-16] = rs1[31-16] + Zext(rs2[15-0])
  * rD[15-0] = rs1[15-0] + Zext(rs2[15-0])
* Is pv.add.sci.h understood as?
  * rD[31-16] = rs1[31-16] + Zext(Imm6[5-0])
  * rD[15-0] = rs1[15-0] + Zext(Imm6[5-0])

### LINKED ISSUE #334: Debug_req pulse width (bug)
If the debug_req input signal is asserted for less than 5 clocks, the core does not properly enter debug mode. Depending on the pulse width, the core may fetch from the debug address (DM_HALT_ADDRESS) but may not the internal debug_mode bit.

The solution is to do one of the following:
1) Document a requirement that the debug_req must be asserted more than 5 clocks (which is reasonable assumption that debug_req is typically asserted for much longer, but may not cover all SOC implementations in the future and can lead to deadlock scenarios if not strictly adhered to)
2) Add a sticky debug_req capture register that is asserted until debug_mode is entered. It should then be cleared when debug_mode==1 and debug_req=0. This is a more robust solution to handle corner cases (especially shown in random verification)

silabs-PaulZ:
Thi PR fixes #334 

The debug_req_i is capture and extended until debug_mode is enterred.


davideschiavone:
Hi Paul, can you please solve these conflicts? It looks good to be to merge once this is done

silabs-PaulZ:
Hi @davideschiavone , I fixed the conflicts.

Issue #131, opened 2019-08-27
Addition of 'not sleeping pin'


Silabs-ArjanB:
The environment of RI5CY cannot judge whether RI5CY still requires its clock (or whether it is for example blocked in a WFI). Addition of a ‘not sleeping’ pin can provide this information. At chip level this information can be used for power management. This feature request amounts to simply making the already existing clock_en signal in riscv_core.sv available on a newly added cpu_sleeping_n pin (see attached pdf).

[RI5CY_Not_Sleeping_Pin.pdf](https://github.com/pulp-platform/riscv/files/3544751/RI5CY_Not_Sleeping_Pin.pdf)

Note: This feature has also been requested for Ibex (https://github.com/lowRISC/ibex/issues/258)

davideschiavone:
Probably you can use core_busy_o.
1 if the core is doing something, 0 otherwise.

arjanbink:
core_busy_o does not provide enough information to judge whether RI5CY needs a clock or not. Suppose we would use (only) core_busy_o to gate clk_i, then there would be no way to wake up from WFI anymore (as there is no combinational path from irq_i to core_busy_o; the same holds for debug_req_i). You would really need to have external access to (or externally duplicate)  RI5CY's clock_en signal. (Alternatively the semantics of core_busy_o could be changed, but of course other people/circuits might depend on its current definition, so it is probably not a good idea to change that).

Silabs-ArjanB:
Fixed with https://github.com/openhwgroup/cv32e40p/pull/391

PR #341, opened 2020-05-15
Added sleep pin indicating when clk_i is allowed to be gated externally

Silabs-ArjanB:
Adde core_sleep_o pin indicating when the clock (clk_i) is allowed (but not required) to be gated externally. The initial state (during and out of reset) of core_sleep_o = 0; the core_sleep_o pin can only possibly be set to 1 in response to executing a WFI instruction.

Once core_sleep_o is 1 it will go to 0 again only for the following conditions:

PULP_CLUSTER = 0: Any pending interrupt or debug request (debug_req_i == 1)
PULP_CLUSTER = 1: clock_en_i == 1

The addition was slightly more complicated than suggested in #131 as the default state of clock_en is 0 during and closely after reset (and we do not want to indicate that the core is asleep during reset assertion nor when just coming out of reset before even an WFI has been executed). 

### DIFF
--- a/rtl/riscv_core.sv
+++ b/rtl/riscv_core.sv
@@ -103,9 +103,9 @@ module riscv_core
   // Debug Interface
   input  logic        debug_req_i,
 
-
   // CPU Control Signals
   input  logic        fetch_enable_i,
+  output logic        core_sleep_o,
   output logic        core_busy_o
 );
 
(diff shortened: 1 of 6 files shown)


Silabs-ArjanB:
@davideschiavone Please hold off on this merge until we have unified the core_busy_o and core_sleep_o behavior.

Silabs-ArjanB:
Superseded by https://github.com/openhwgroup/cv32e40p/pull/391

PR #349, opened 2020-05-19
Renamed riscv_core to cv32e40p_core

Silabs-ArjanB:
RTL update for https://github.com/openhwgroup/core-v-docs/issues/51

Only the module renaming was done. The other topics raised (e.g. license, ifdefs) are not related to documentation and will be handled separately.

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -30,7 +30,7 @@ sources:
     - rtl/riscv_pmp.sv
     - rtl/riscv_prefetch_buffer.sv
     - rtl/riscv_prefetch_L0_buffer.sv
-    - rtl/riscv_core.sv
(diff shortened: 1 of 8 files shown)


davideschiavone:
Hi @Silabs-ArjanB ,

I can somehow modify @silabs-PaulZ PRs by solving conflicts, but I cannot with yours.
Any idea? probably some permissions.
How would you like to deal with that?
To me it's fine, we should fork and make PR to your branch, and then you accept them to solve the conflix. 

Let me know what you think about it

Silabs-ArjanB:
Hi @davideschiavone I don't know what causes that; maybe we all have a different way of working in git, but as far as I know Paul and I are following the same flow. Please do what you proposed to see if that will work.

PR #350, opened 2020-05-22
Fix linter warnings

wallento:
Fix some implicit variables and unconnected pins.

Signed-off-by: Stefan Wallentowitz <stefan.wallentowitz@hm.edu>

PR #352, opened 2020-05-23
Add FuseSoC support

wallento:
Hi,
this is a draft PR that adds FuseSoC support. Before merging we need:

- [ ] Upstream fpnew core file, ideally in its own repository. That also removes the dependency to clone the repo at a particular path. Before this can happen, we need a minor fix to FuseSoC for include path handling (@olofk is working on it)
- [ ] Support for the `core` simulation (working on it right now)
- [ ] Support for synthesis targets if people agree thats a good way to move forward. A lot of things could be adopted from the ibex repository, that did some groundbreaking improvements to FuseSoC and use many of the features.

Cheers,
Stefan


wallento:
Depends on #351 and #350

PR #367, opened 2020-06-16
Fusesoc and linter

wallento:


tb/core/tb_top_verilator.cpp
```diff
@@ -69,7 +69,7 @@ void dump_memory()
 {
     errno = 0;
     std::ofstream mem_file;
-    svLogicVecVal addr = {0};
+    svLogicVecVal addr = {0, 0};
```
MikeOpenHWGroup, line 72: Why is this necessary?  Is it related to using fusesoc, or is it 'cleanup'?

tb/core/tb_top_verilator.cpp
```diff
@@ -84,7 +84,7 @@ void dump_memory()
 
         std::cout << "finished dumping memory" << std::endl;
 
-    } catch (std::ofstream::failure e) {
+    } catch (std::ofstream::failure &e) {
```
MikeOpenHWGroup, line 87: As as comment above.

tb/core/tb_top_verilator.cpp
wallento: That is a cleanup to make newer GCC happy.

tb/core/tb_top_verilator.cpp
wallento: Also GCC issues a warning if you use an instance not a reference.

wallento:
Thanks, sorry, I have closed this one, because it was an accident (before testing). Will copy the comments over to the new one (#368)

MikeOpenHWGroup:
Never be sorry for issuing a pull-request.  We are pleased to have your contribution (even if we choose not to merge it in).

PR #368, opened 2020-06-16
Fusesoc and Linter

wallento:
This adds support for FuseSoC. As example, it fixes the verilator model to directly use it.
But even more interesting, it adds a github action that automatically runs the linter on each push and pull request. When the linter fails, the action fails and a PR can be polished.
As a caveat, there are 421 Linter warnings on the codebase at the moment. I have added a waiver file that suppresses them, but I suggest to fix the issues or replace the placeholder comments with an explanation. Happy to help, but too many for a "quick PR".


wallento:
Here is an example for the annotations: https://github.com/wallento/riscv/actions/runs/137172924
At an PR it will put them right to the code changes.

wallento:
From #367:
> Thanks for getting the ball rolling on this @wallento. I've added a few minor comments, and have three big ones:
> 
>     1. A lot of the "tb" dirs/files in this repo are slated to be removed to encoruage people to use [core-v-verif](https://github.com/openhwgroup/core-v-verif) testbenches, so we'll need to replicate some of this there.
> 
>     2. tb/verilator-model/Makefile has a complete manifest embedded within it.  This is exactly what we are trying to displace.
> 
>     3. Many (all?) of the RTL port map and parameter changes you suggest have been (or are being) taken care of elsewhere.

The minor things are to resolve warnings with recent GCCs.

About 2, you are right. I kept it in there for reference. If you are happy I would replace it with `Makefile.fusesoc` in that same folder. It does the same job. I would suggest, but didn't want to be too drastic :)
About 3, you are right, seems it slipped in there and relates to #350 that I fixed while doing it. Will look into it and amend.
Thanks!

wallento:
Hi @MikeOpenHWGroup,
I have ameneded the PR. The changes of 3 are necessary to make the verilator-model run again. I believe long term it would be more useful to have a more sophisticated verilator-model as Quick Start vs. this rather simple bench and the full blown verification flow. I would be happy to contribute this as I am pretty fluent in it. I will also add a CI test to it, if you want.
I have remove the original Makefile, this change pretty much shows the power FuseSoC :)
Best,
Stefan

wallento:
I can't really understand what Github feels about that Makefile to be a conflict, seems pretty straight-forward to me..

wallento:
Ah, I see! There was a massive rename between my first PR and now :) I will amend again after rebase!

bluewww:
PR is guaranteed to be broken with this merge https://github.com/openhwgroup/cv32e40p/pull/360#event-3449965290. I believe a rebase is in order

wallento:
Rebased and amended to match the changed naming

.github/workflows/lint.yml
```diff
@@ -0,0 +1,18 @@
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Lint Verilog sources with Verilator
```
Silabs-ArjanB, line 9: I would not want any dependency on Verilator

.github/workflows/lint.yml
```diff
@@ -0,0 +1,18 @@
...
+        uses: librecores/ci-fusesoc-action@master
+        with:
+          command: 'run'
+          core: 'openhw:cv32e40p:core'
+          target: 'lint'
+          tool: 'verilator'
```
Silabs-ArjanB, line 15: I would not want any dependency on Verilator

cv32e40p.core
```diff
@@ -0,0 +1,79 @@
...
+    file_type: systemVerilogSource
+
+targets:
+  default: &default
+    filesets:
+      - tool_verilator ? (files_lint_verilator)
+      - files_rtl
+  lint:
+    <<: *default
+    default_tool: verilator
```
Silabs-ArjanB, line 73: Can we make sure things run without Verilator as well?

Silabs-ArjanB:
Hi @wallento Is there a way we can merge your pull request and also have the possibility to keep the old way of working for 1-2 weeks so that we do not immediately block many people just in case there are initial problems? How about core-v-verif, that would need to be updated as well, right?

PR #370, opened 2020-06-17
improved tb for interrupt tests

davideschiavone:
Added random stalls in interrupts
Stopping simulation when PULP_OBI=0 is used in RANDOM Stalls as it's not supported yet

### DIFF
--- a/tb/core/interrupt_test/interrupt_test.c
+++ b/tb/core/interrupt_test/interrupt_test.c
@@ -515,8 +515,55 @@ void mat_mult(uint32_t mat1[MAT_DIM][MAT_DIM], uint32_t mat2[MAT_DIM][MAT_DIM], 
     }
 }
 
+void activate_random_stall(void)
+{
+  // Address vector for rnd_stall_reg, to control memory stalls/interrupt
+  volatile unsigned int *rnd_stall_reg[16];
+
+  // Setup the address vector
+  rnd_stall_reg[0] = 0x16000000;
+  for (int i = 1; i < 16; i++) {
+    rnd_stall_reg[i] = rnd_stall_reg[i-1] + 1; // It is a pointer to int ("+ 1" means "the next int")
+  }
+
+  /* The interposition of the stall generator between CPU and MEM should happen BEFORE the stall generetor is active */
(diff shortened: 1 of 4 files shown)


tb/core/interrupt_test/interrupt_test.c
```diff
@@ -515,8 +515,55 @@ void mat_mult(uint32_t mat1[MAT_DIM][MAT_DIM], uint32_t mat2[MAT_DIM][MAT_DIM],
     }
 }
 
+void activate_random_stall(void)
```
Silabs-ArjanB, line 518: Could we move this function into a separate file maybe? Now this function differs from the one used in https://github.com/davideschiavone/riscv/blob/improved_tb/tb/core/custom/hello_world.c (specifically the order of enabling is different).

If the Hello World program will be removed then you can ignore my comment.

tb/core/mm_ram.sv
```diff
@@ -294,6 +294,7 @@ module mm_ram
 
                 // write to rnd stall regs
                 end else if (data_addr_i[31:16] == 16'h1600) begin
+                    if(~is_pulp_obi) begin $fatal("[RANDOM STALL] Not SUPPORTED RANDOM STALL FOR OBI at the moment"); $stop; end
```
Silabs-ArjanB, line 297: In https://github.com/Silabs-ArjanB/cv32e40p/tree/ArjanB_obi_fix_base I made sure that random stalls were supported for OBI. Did you end up not merging those test bench changes? What is your strategy for this? If you are going to support random wait states on gnt and rvalid in the future, then I am fine with not supporting it at the moment. If the intention is to not support it, then I think we should also not support it for any configuration. Of course, no matter what, random wait states need to be supported in core-v-verif.

tb/core/interrupt_test/interrupt_test.c
davideschiavone: done in another PR

tb/core/mm_ram.sv
davideschiavone: There was a bug fixed in another PR

davideschiavone:
Closing as this has been taken over by PR #380

PR #385, opened 2020-06-24
Add `tinfo` debug CSR

accuminium:
The [RISC-V debug specification, Section 5.2.5](https://github.com/riscv/riscv-debug-spec/raw/release/riscv-debug-release.pdf) defines the *Trigger Info* register, which returns the trigger types supported by the currently selected trigger.  CV32E40P implements *address/data match* triggers (or no triggers if `DEBUG_TRIGGER_EN` is disabled), and this PR exposes the supported trigger in a newly added `tinfo` CSR.

### DIFF
--- a/rtl/cv32e40p_cs_registers.sv
+++ b/rtl/cv32e40p_cs_registers.sv
@@ -254,9 +254,9 @@ module cv32e40p_cs_registers
   logic [31:0] mepc_q, mepc_n;
   logic [31:0] uepc_q, uepc_n;
   // Trigger
-  logic [ 3:0] tmatch_control_type;
   logic [31:0] tmatch_control_rdata;
   logic [31:0] tmatch_value_rdata;
+  logic [15:0] tinfo_types;
   // Debug
   Dcsr_t       dcsr_q, dcsr_n;
   logic [31:0] depc_q, depc_n;
@@ -421,7 +421,7 @@ if(PULP_SECURE==1) begin
       CSR_TDATA2:
                csr_rdata_int = tmatch_value_rdata;
       CSR_TINFO:
-               csr_rdata_int = 1'b1 << tmatch_control_type;
+               csr_rdata_int = tinfo_types;
 
       CSR_DCSR:
                csr_rdata_int = dcsr_q;//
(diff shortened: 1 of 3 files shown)


Silabs-ArjanB:
Hi @accuminium, 

Thank you very much for this pull request! 

Could you please also make the corresponding update in the CSR chapter of the User Manual and also mention the new CSR in the Debug chapter (see https://core-v-docs-verif-strat.readthedocs.io/projects/cv32e40p_um/en/latest/control_status_registers.html and https://core-v-docs-verif-strat.readthedocs.io/projects/cv32e40p_um/en/latest/debug.html).

The source files for the documentation can be found in https://github.com/openhwgroup/core-v-docs/tree/master/cores/cv32e40p/user_manual/source

Silabs-ArjanB:
@accuminium Perfect, thanks!

silabs-PaulZ:
Hi @accuminium ,

Since the `type` cannot be changed, reading the `tinfo` register will return an illegal_insn. This is also how IBEX implemented it.

The debug specification has the following statement about TINFO:
image

The spec does say 'may be unimplemented'. So, I agree that we should make a change to implement this register which will ease the debugger usage.

Even though your implementation works, can you consider slightly changing it to ensure future changes are less prone to bugs. When reading `tinfo`, you are basing the result on the active value of `type `instead of reflecting all valid types. By changing the logic now, it will hopefully prevent the wrong implementation in the future. BTW, we are debating adding etrigger type.

My suggestion:
Create the following enumeration in cv32e40p_defines.sv
```
typedef enum logic[3:0] {
TTYPE_MCONTROL = 4’h2,
TTYPE_ICOUNT = 4’h3,
TTYPE_ITRIGGER = 4’h4,
TTYPE_ETRIGGER = 4’h5
} trigger_type_e
```

In the cv32e40p_cs_registers.sv, make a logic variable containing the valid type(s) and use that when reading `tinfo `(decoupling `type`).
```
logic [3:0] tinfo_types;
...
assign tinfo_types = 1 << TTYPE_MCONTROL;
...
CSR_TINFO:
   csr_rdata_int = tinfo_types;
```

accuminium:
Thanks for your feedback, @silabs-PaulZ, I have implemented it in two new commits.

Issue #384, opened 2020-06-23
Do not make glob imports of packages in the compilation unit scope unless necessary


bluewww:
There are imports like this 
https://github.com/openhwgroup/cv32e40p/blob/8e22c994198ad08f4cacf239d56d98e1cc25627b/rtl/cv32e40p_core.sv#L30-L38

which makes all the variables visible in the whole compilation unit. This can cause issues when this codebase is compiled in the same compilation unit as another module that has a variable with the same name as in one of those packages. We saw this issue when using vcs and vsim. In vsim we had one compilation unit per file, but vcs had one compilation unit for the whole design.

The better approach is doing the import inside the module or in the module header (same effect).

i.e. 
```SystemVerilog
module cv32e40p_core import cv32e40p_defines::*;
#(
  parameter PULP_HWLP           =  0,                   // Hardware Loop (not supported yet; will be supported)
  parameter PULP_CLUSTER        =  0,

[...]
```

Silabs-ArjanB:
Hi @bluewww Fully agreed; could you actually make a pull request for these changes yourself?

bluewww:
Sorry I didn't get around to make those changes, but I see you already worked on it. Thanks.

Silabs-ArjanB:
Hi @bluewww The corresponding pull request has been merged. Could you please close this issue if you agree it has been fixed or let us know what issue still exist?

bluewww:
@Silabs-ArjanB thanks looks correct

Issue #301, opened 2020-04-08
Use of SV compile macros in RTL


MikeOpenHWGroup:
The RTL code for cv32e40p contains a large number of compile-time macro checks.  For example, in [riscv_pmp.sv](https://github.com/openhwgroup/cv32e40p/blob/master/rtl/riscv_pmp.sv), lines 233..242 we see:
```
`ifdef EN_NAPOT_RULE_8B
       `RULE_0:
       begin: BYTE_ALIGN_8B
          mask_addr[i]  = 32'hFFFF_FFFE;
          start_addr[i] = pmp_addr_i[i] & mask_addr[i];
          stop_addr[i]  = pmp_addr_i[i];

          `ifdef DEBUG_RULE $display(" NAPOT[%d]  --> BYTE_ALIGN_8B: %8h<-- addr --> %8h", i, start_addr[i]<<2, stop_addr[i]<<2 ); `endif
       end
`endif
```
There are three issues highlighted here:
1. Macros like `EN_NAPOT_RULE_8B` are defined in the file that uses them.  These should be moved to a central location for all such macros.
2. Macros likes `RULE_0` are used as constants.  However, since macros have global scope, this could affect the entire design, so it is recommended to define these as localparams.
3. Macros are global, so things like `DEBUG_RULE` could create very unexpected behavior if it is used in another way elsewhere in the RTL.

I would like to see the following:
1. Prefix all macros with CV32E40 (e.g. `CV32E40P_EN_NAPOT`) to maximize the likelihood of having unique macros. 
2. Move all macro definitions (\`define) to a central file.
3. Update the manifest to ensure that the macro definitions are read in first to ensure that all simulation, formal verification and synthesis are using the same model.
4. Replace macros used as constants with localparams.

Attached are two files: `rtl_ifdefs.uniq.txt` lists all the unique uses of macros in the RTL and `rtl_ifdefs.uniq.txt` shows where these are defined.
[rtl_ifdefs.uniq.txt](https://github.com/openhwgroup/cv32e40p/files/4452194/rtl_ifdefs.uniq.txt)
[rtl-defines.txt](https://github.com/openhwgroup/cv32e40p/files/4452195/rtl-defines.txt)

bluewww:
I agree with all the points you have brought up, especially the namespacing is useful. This is a common pitfall in the C language too.

[...]

PR #389, opened 2020-06-29
Arjan b pkg import

Silabs-ArjanB:
Fix for #384 and #301 

Furtermore I cleaned up the inconsistent and undesired dependence on defines. The following defines were assumed in the original code:

- SYNTHESIS
- VERILATOR
- APU_TRACE
- TRACE_EXECUTION
- PULP_FPGA_EMUL

SYNTHESIS was used for

- guarding non-synthesizble RTL (clock gate)
- guarding $display statements
- guarding $fatal statements
- guarding assertions

VERILATOR was used for:

- guarding (non-synthesizable) cv32e40p_tracer
- guarding assertions

APU_TRACE was used for:

- APU trace generation

TRACE_EXECUTION was used for:

- trace generation

Now all this has been removed (except for the guard around assertions which will be removed soon as well as the assertions will move out of the RTL files).

The following was done to remove the dependency on defines and to not have behavioral code inside the RTL:

- Moved behavioral tracer file and its package out of rtl directory to tb/core/cv32e40p_tracer.sv and tb/core/include/cv32e40p_tracer_pkg.sv respectively
- Moved behavioral clock gate file out of rtl directory to (new) bhv directory
- Moved $display parts out of RTL files into testbench (new file tb/core/cv32e40p_core_log.sv)
- Moved APU trace part out of RTL into testbench (new file tb/core/cv32e40p_apu_tracer.sv)
- Moved tracer instantiation out of RTL into testbench (notice that also Ibex moved this behavioral code out of there top level RTL)
- Renamed TRACE_EXECUTION to CV32E40P_TRACE_EXECUTION
- Renamed APU_TRACE to CV32E40P_APU_TRACE
- Both FPGA and ASIC version of clock gate are now handled via manifest file (removed custom ifdefs from cv32e40p_sim_clock_gate.sv) (originally one target was handled via ifdef, the other one via manifest)
- Guarded assertions by `ifdef ASSERT_ON (and no longer by `ifndef VERILATOR or `ifndef SYNTHESIS). The verification team is going to move assertions to separate files our of the rtl directory, after which also this final define will have been removed from the RTL (they can rename ASSERT_ON to something else and/or put the ifdefs around the bind commands instead) .
- defines by localparam in various places


davideschiavone:
Hi @Silabs-ArjanB , why did you removed the pmp file? What happens if someone wants to fork and implement the PMP by themself starting from our RTL? I suggest to keep the file.

Silabs-ArjanB:
Hi @davideschiavone I removed it because it is full of defines and I didn't want to spend time cleaning that up. Also there are various other known bugs with PMP and also PMP is not currently integrated properly in IF or LSU. My idea was to have a good look at the Ibex PMP (which recoded it to a much more condensed version than our RTL) when we are at the point that we want to integrate the functionality again.

Silabs-ArjanB:
Corresponding doc update is https://github.com/openhwgroup/core-v-docs/pull/120

tb/core/cv32e40p_tracer.sv
```diff
@@ -774,11 +767,11 @@ module cv32e40p_tracer import cv32e40p_pkg::*;
     $fclose(f);
   end
 
-  assign rd  = {rd_is_fp,  instr[`CV32E40P_REG_D]};
-  assign rs1 = {rs1_is_fp, instr[`CV32E40P_REG_S1]};
-  assign rs2 = {rs2_is_fp, instr[`CV32E40P_REG_S2]};
-  assign rs3 = {rs3_is_fp, instr[`CV32E40P_REG_S3]};
-  assign rs4 = {rs3_is_fp, instr[`CV32E40P_REG_S4]};
+  assign rd  = {rd_is_fp,  instr[11:07]};
```
davideschiavone, line 769: Why aren't we using the local PARAMETERS here? @Silabs-ArjanB

davideschiavone:
Hi @Silabs-ArjanB , if you can solve conflicts I can try it out on PULP as well to check that nothing has been broken and then merge. Due to the amount of file changed, I won't accept any PR before merging this one.

tb/core/cv32e40p_tracer.sv
Silabs-ArjanB: Simply because these numbers are fixed due to the ISA and if we would introduce local params here we would only use them once

Silabs-ArjanB:
Hi @davideschiavone I did a merge from master. If you want me to add localparams in the tracer file please let me know and I will do that.

Issue #379, opened 2020-06-22
Cannot gate clock for WFI when PULP_CLUSTER = 1


Silabs-ArjanB:
In the configuration with PULP_CLUSTER = 1 the internal architectural clock gate (i.e. core_clock_gate_i) cannot be gated.

Once the WFI has been hit, the core_busy_o signal will go low as intended. In order to gate the internal clock the environment will have to lower clock_en_i as the final gating condition is determined as follows:

  assign clock_en      = PULP_CLUSTER ? clock_en_i | core_busy_o : irq_pending | debug_req_i | core_busy_o;

Once both clock_en_i and core_busy_o are low a pending and locally enabled interrupt will not be signaled via core_busy_o as clock_en will remain low and the core will not become 'busy'. The environment could have kept clock_en_i high to prevent this situation, but that would leave the internal clocks running during a WFI, which is not desired. 

Likely irq_pending (and debug_req_i ?) need to factor into clock_en for PULP_CLUSTER = 1 as well.

davideschiavone:
The WFI should not be used when the PULP_CLUSTER parameter is set to 1. 
I suggest sto implement the WFI as a NOP in that case and to create a new documentation issue to do so

davideschiavone:
Separatly in another response here, the even unit handles the interrupts and debug requests, i.e., it receives them and forwards them to the core after waking it up. That's why the are not part of the formula of clock_en. When the core uses the PULP_CLUSTER=1 it relies complitly on the event unit for power managment.

Silabs-ArjanB:
Okay, that makes sense. So the CV32E40P with PULP_CLUSTER = 1 will never receive an interrupt nor a debug request while pulp_clock_en_i = 0, right?

Silabs-ArjanB:
Can we define the requirements for PULP_CLUSTER = 1 as follows?

- WFI shall be implemented as NOP (such that it does not impact core_sleep_o)
- While pulp_clock_en_i = 0, the environment must guarantee that instr_gnt_i = 0, instr_rvalid_i = 0, data_gnt_i = 0, data_rvalid_i = 0, debug_req_i = 0, irq_i[] = 0 (not sure if we should demand this independent of local/global interrupt enable bits or not)
- pulp_clock_en_i = 0 is only allowed when core_sleep_o = 1

PR #391, opened 2020-07-01
Added/corrected sleep interface; fixed #373, #362, #379

### LINKED ISSUE #373: p.elw rD, Imm(rs1!) should not exist
Currently two flavor of p.elw are in the RTL:

    p.elw rD, Imm(rs1!)
    p.elw rD, Imm(rs1)

The encoding that currently is interpreted as p.elw rD, Imm(rs1!) should trigger an illegal instruction instead. See also https://github.com/openhwgroup/core-v-docs/issues/99

Silabs-ArjanB:
Added/corrected sleep interface:

- Replacement of core_busy_o with core_sleep_o
- If PULP_CLUSTER = 0, then core_sleep_o can only become asserted in response to a WFI instruction
- If PULP_CLUSTER = 1, then core_sleep_o can only become asserted in response to a P.ELW instruction
- Keeping internal clock gated until fetch_enable_i sampled high
- Made WFI act as true NOP for PULP_CLUSTER = 1 (i.e. sleep state is never entered)
- Renamed clock_en_i to pulp_clock_en_i
- Made p.elw an illegal instruction for PULP_CLUSTER = 0
- Moved sleep related circuitry into its own file (cv32e40p_sleep_unit.sv)

Fixed issues: #373 #362 #379 

### DIFF
--- a/rtl/cv32e40p_core.sv
+++ b/rtl/cv32e40p_core.sv
@@ -47,8 +47,8 @@ module cv32e40p_core
   input  logic        clk_i,
   input  logic        rst_ni,
 
-  input  logic        clock_en_i,    // enable clock, otherwise it is gated
-  input  logic        scan_cg_en_i,  // enable all clock gates for testing
+  input  logic        pulp_clock_en_i,                  // PULP clock enable (only used if PULP_CLUSTER = 1)
+  input  logic        scan_cg_en_i,                     // Enable all clock gates for testing
 
   // Core ID, Cluster ID, debug mode halt address and boot address are considered more or less static
   input  logic [31:0] boot_addr_i,
@@ -98,7 +98,7 @@ module cv32e40p_core
 
   // CPU Control Signals
   input  logic        fetch_enable_i,
-  output logic        core_busy_o
+  output logic        core_sleep_o
 );
 
   // Unused parameters and signals (left in code for future design extensions)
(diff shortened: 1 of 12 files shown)


rtl/cv32e40p_core.sv
```diff
@@ -1237,42 +1237,46 @@ module cv32e40p_core
...
+  // Assumptions
   //----------------------------------------------------------------------------
 
-`ifndef VERILATOR
-
-  // Check that IRQ indices which are reserved by the RISC-V privileged spec 
+  // Assume that IRQ indices which are reserved by the RISC-V privileged spec 
   // or are meant for User or Hypervisor mode are not used (i.e. tied to 0)
   property p_no_reserved_irq;
      @(posedge clk_i) disable iff (!rst_ni) (1'b1) |-> ((irq_i[15:12] == 4'b0) && (irq_i[10:8] == 3'b0) && (irq_i[6:4] == 3'b0) && (irq_i[2:0] == 3'b0));
```
davideschiavone, line 1249: Can we please remove this assertion and add it back once we do the interrupt localparameter mask?

we can do it directly in this PR or preferably take care of interrupts in a separate PR

@Silabs-ArjanB

Silabs-ArjanB: Agreed; done

Issue #404, opened 2020-07-10
Missed debug_req_i pulse


Silabs-ArjanB:
A debug_req_i pulse can be missed if it comes in during the part of the enable process of CV32E40P in which the interal clock is gated off.

MikeOpenHWGroup:
Hi @Silabs-ArjanB - good catch.  I assume you caught that with inspection, rather than simulation.  What would a testbench need to do to ensure it could reproduce this issue.  Are there similar issues that verification should be aware of?

Silabs-ArjanB:
This was the result of a directed formal test I did aimed at finding issues with the architected clock gate. The statement that a 1 cycle pulse on debug_req_i is enough is part of the User Manual.

MikeOpenHWGroup:
Nice.   I've created [a Vplan issue](https://github.com/openhwgroup/core-v-docs/issues/128) for @silabs-oysteink to ensure we capture that in the Vplan.

I suspect there is more to this issue: your formal test looks for one cycles pulses on debug_req_i, but the issue seems to be related to "part of the enable process".  What is that?

Silabs-ArjanB:
With the enable process I mean the assertion of fetch_enable_i as described in https://core-v-docs-verif-strat.readthedocs.io/projects/cv32e40p_um/en/latest/integration.html. Certain other signals depend on that signal (as described in the same section) and clock gating depends on it as well (as also described but awaiting a pull request)

PR #405, opened 2020-07-10
Fix for #404

Silabs-ArjanB:


jm4rtin:
``clk`` and ``clk_i`` for gated and ungated clock is quite confusing as ``_i`` indicates an input but they are both inputs. What about something like ``gclk_i`` and ``clk_i``? Maybe such a refactor should be a separate PR.

Silabs-ArjanB:
@jm4rtin Would you be okay for clk_ungated for the free running clock (except of course for the top level pin which would keep its clk_i name)? Otherwise we have to change all files

jm4rtin:
@Silabs-ArjanB Something like that could be okay. I agree changing all of the RTL would be painful.

PR #410, opened 2020-07-14
:bug: fix naming conflict

mp-17:
Dear All,

From the SystemVerilog specs, in the section for the conditional-generate statement, we read:

"_It is not permissible for any of the named generate blocks to have the same name as any other
declaration in the same scope, even if that block is not selected for instantiation._"

In the **cv32e40p_sleep_unit**, one of the conditional-generate statements is named **SLEEP**, but this name already exist in the imported package.
This causes an error to occur during the synthesis.

To solve this problem, I changed **SLEEP** to **NO_PULP_SLEEP**.

Best regards,
Matteo

### DIFF
--- a/rtl/cv32e40p_sleep_unit.sv
+++ b/rtl/cv32e40p_sleep_unit.sv
@@ -1,5 +1,5 @@
 // Copyright 2020 Silicon Labs, Inc.
-//   
+//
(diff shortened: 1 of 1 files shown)


davideschiavone:
For me looks good, @Silabs-ArjanB ?

PR #413, opened 2020-07-16
Added cv32e40p wrapper

davideschiavone:
This file instatiates both the core and the tracer as it was before.
Users that want to use the tracer should simply replace in their system the core_i instantiation with the wrapper_i one.

I also suggest that the core-verif repository uses such wrapper to avoid dealing with the core internal signals,
otheriwse any internal modification of the core must be ported to the core-verif and any other system that uses the tracer.

The bhv/cv32e40p_wrapper must be supported as much as any other cv32e40p rtl file.

### DIFF
--- a/cv32e40p_manifest.flist
+++ b/cv32e40p_manifest.flist
@@ -58,3 +58,6 @@ ${DESIGN_RTL_DIR}/cv32e40p_sleep_unit.sv
 ${DESIGN_RTL_DIR}/cv32e40p_core.sv
 ${DESIGN_RTL_DIR}/../bhv/cv32e40p_sim_clock_gate.sv
 ${DESIGN_RTL_DIR}/../bhv/cv32e40p_tracer.sv
+${DESIGN_RTL_DIR}/../bhv/cv32e40p_wrapper.sv
+${DESIGN_RTL_DIR}/../bhv/cv32e40p_core_log.sv
+${DESIGN_RTL_DIR}/../bhv/cv32e40p_apu_tracer.sv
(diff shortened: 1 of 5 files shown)


example_tb/core/cv32e40p_tb_subsystem.sv
```diff
@@ -60,81 +60,15 @@ module cv32e40p_wrapper
...
-    cv32e40p_core
+    cv32e40p_wrapper
         #(
           .PULP_XPULP            ( 1                     ),
           .PULP_CLUSTER          ( PULP_CLUSTER          ),
           .FPU                   ( FPU                   ),
           .PULP_ZFINX            ( PULP_ZFINX            ),
           .NUM_MHPMCOUNTERS      ( 1                     ))
-    core_i
+    core_wrapper_i
```
Silabs-ArjanB, line 71: Can you rename this to wrapper_i? I have tried in the RTL itself to always use the naming style 'cv32e40p_xyz xyz_i'

davideschiavone: sure, done

davideschiavone:
> 
> 
> Hi Davide, looks good. I am only requesting a minor renaming; other than that good to go!

I have done it thanks

davideschiavone:
> 
> 
> One more remark: The manifest file needs to be adapted as well and core-v-verif will need to be updated as well; that was the reason for this change.

I have updated the manifest file as well now

cv32e40p_manifest.flist
```diff
@@ -58,3 +58,4 @@ ${DESIGN_RTL_DIR}/cv32e40p_sleep_unit.sv
 ${DESIGN_RTL_DIR}/cv32e40p_core.sv
 ${DESIGN_RTL_DIR}/../bhv/cv32e40p_sim_clock_gate.sv
 ${DESIGN_RTL_DIR}/../bhv/cv32e40p_tracer.sv
```
Silabs-ArjanB, line 60: The other bhv files will need to be put in this list as well now (apu tracer and log)

davideschiavone: I force pushed this change as well

PR #422, opened 2020-07-20
Conditionally compile tracer

MikeOpenHWGroup:
A couple of conditionally compiler macros.  These are necessary for the CORE-V-VERIF `core` (verilator) testbench.

### DIFF
--- a/bhv/cv32e40p_apu_tracer.sv
+++ b/bhv/cv32e40p_apu_tracer.sv
@@ -35,6 +35,8 @@
 //                                                                            //
 ////////////////////////////////////////////////////////////////////////////////
 
+`ifdef CV32E40P_APU_TRACE
+
 module cv32e40p_apu_tracer
 (
 );
@@ -72,3 +74,5 @@ module cv32e40p_apu_tracer
      end
 
 endmodule // cv32e40p_apu_tracer
+
+`endif // CV32E40P_APU_TRACE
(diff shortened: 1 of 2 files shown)


Silabs-ArjanB:
Hi @MikeOpenHWGroup If we need to temporarily have these ifdefs such that you can make progress in core-v-verif, then that is okay, but in the end we do not want any ifdef/ifndef in these files whatsoever. Inclusion should eventually be proeprly fixed either in the wrapper and/or the manifest file. So, if you want we can accept this pull request as a temporary solution.

MikeOpenHWGroup:
@Silabs-ArjanB I certainly agree that getting rid of ifdef/ifndef for conditional compile should the goal.  I believe we will need a manifest generator to make that happen given our desire to continue to supporting Verilator.  So I very much appreciate the offer to accept this PR and understand it is a temporary solution.

My primary interest here is establishing a prototype flow to extract the Assertions extracted from RTL and making the Assertions visible to the UVM environment.

If there is anything I can do to hurry things along toward a final and proper solution, please let me know!

MikeOpenHWGroup:
> Approving for now. 

Thanks @Silabs-ArjanB.  Who executes the merge?

Silabs-ArjanB:
> 
> 
> > Approving for now.
> 
> Thanks @Silabs-ArjanB. Who executes the merge?

@davideschiavone

PR #437, opened 2020-08-06
[fpga] Modify Perf. Counter for DSP Inference

ganoam:
This commit modifies performance counter implementation to allow
inference of DSP slices to absorb counter logic on FPGAs.
By making use of the dedicated circuitry we are able to save both logic
and flip-flop resources.

Summary of changes:
- Counter logic is moved to a new separate module
- The preprocessor variable `TARGET_XILINX` (automatically set by
  Bender) is used to include the required synthesis pragma.
- DSP Inferenceis supported for Xilinx FPGA devices featureing DSP48E1
  slices or similar and for counter widths up to 48 bits. However, for
  widths of <= 32 bits only, the corresponding flip-flops can be
  absorbed into the DSP's output pipeline register.
- Therefore, a new top-level parameter `WIDTH_MHPMCOUNTERS` is
  introduced to control the width of the *additional* performance
  counters.
- The two primary counters `mcycle` and `minstret` are unaffected and
  remain at a width of 64 bits and are implemented in logic fabric.

Concrete savings: (Synthesis with 29 32-bit counters, target: Virtex-7)
```
                LUT     FF      DSP
 ----------------------------------
 baseline:      8951    3858    7
 opt:           7976    2917    36
 ----------------------------------
 Diff           -975    -941    +29
                -10.9%  -24.4%
```

Signed-off-by: ganoam <gnoam@live.com>

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -16,6 +16,7 @@ sources:
     - rtl/cv32e40p_alu_div.sv
     - rtl/cv32e40p_compressed_decoder.sv
     - rtl/cv32e40p_controller.sv
+    - rtl/cv32e40p_counter.sv
     - rtl/cv32e40p_cs_registers.sv
     - rtl/cv32e40p_decoder.sv
     - rtl/cv32e40p_int_controller.sv
--- a/example_tb/core/Makefile
+++ b/example_tb/core/Makefile
@@ -198,6 +198,34 @@ custom-vsim-run-gui: vsim-all custom/hello_world.hex
 custom-vsim-run-gui: ALL_VSIM_FLAGS += "+firmware=custom/hello_world.hex"
 custom-vsim-run-gui: vsim-run-gui
 
+#Perfcnt test
+custom/perfcnt.elf: custom/perfcnt.c
+	$(RISCV_EXE_PREFIX)gcc -march=rv32imc -o $@ -w -Os -g -nostdlib \
+		$(CUSTOM_GCC_FLAGS) \
+		-T custom/link.ld \
+		-static \
+		custom/crt0.S \
+		$^ mem_stall/mem_stall.c custom/syscalls.c custom/vectors.S \
+		-I $(RISCV)/riscv32-unknown-elf/include \
+		-I mem_stall \
+		-L $(RISCV)/riscv32-unknown-elf/lib \
+		-lc -lm -lgcc
+
+custom/perfcnt.dump: custom/perfcnt.elf
+	$(RISCV_EXE_PREFIX)objdump --disassemble-all $^ > $@
+
+perfcnt-clean:
+	rm -rf custom/perfcnt.elf custom/perfcnt.hex custom/perfcnt.dump
+.PHONY: perfcnt-test
+perfcnt-test-vsim-run: vsim-all custom/perfcnt.hex
+perfcnt-test-vsim-run: ALL_VSIM_FLAGS += "+firmware=custom/perfcnt.hex"
+perfcnt-test-vsim-run: vsim-run
+
+.PHONY: perfcnt-test-vsim-run-gui
+perfcnt-test-vsim-run-gui: vsim-all custom/perfcnt.hex custom/perfcnt.dump
+perfcnt-test-vsim-run-gui: ALL_VSIM_FLAGS += "+firmware=custom/perfcnt.hex"
+perfcnt-test-vsim-run-gui: vsim-run-gui
+
 # Running custom fp programs:
 # This is an example for running a floating-point matrix multiplication in the testbench
 # We link with our custom crt0.s and syscalls.c against newlib so that we can
@@ -254,4 +282,4 @@ interrupt-vsim-run-gui: vsim-run-gui
 
 # general targets
 .PHONY: clean
-clean: tb-clean tb-clean-fp custom-clean custom-fp-clean interrupt-clean
+clean: tb-clean tb-clean-fp custom-clean custom-fp-clean interrupt-clean perfcnt-clean
--- a/example_tb/core/custom/perfcnt.c
+++ b/example_tb/core/custom/perfcnt.c
@@ -1,0 +1,266 @@
+/*
+**
+** Copyright 2020 OpenHW Group
+**
+** Licensed under the Solderpad Hardware Licence, Version 2.0 (the "License");
+** you may not use this file except in compliance with the License.
+** You may obtain a copy of the License at
+**
+**     https://solderpad.org/licenses/
+**
+** Unless required by applicable law or agreed to in writing, software
+** distributed under the License is distributed on an "AS IS" BASIS,
+** WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+** See the License for the specific language governing permissions and
+** limitations under the License.
+**
+*******************************************************************************
+**
+** Performance counter Sanity check
+**
+** Enables performance counters while executing a simple fibonacci test program.
+** Read out the values and perform some simple sanity checks.
+**
+** Make sure to instantiate cv32e40p_wrapper with the parameter
+** NUM_MHPMCOUNTERS = 29. (in core/cv32e40p_tb_subsystem.sv)
+**
+*******************************************************************************
+*/
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+static int check(unsigned long long is, unsigned long long should)
+{
+  int err;
+  err = is == should ? 0 : 1;
+  if (err)
+    printf("fail\n");
+  else
+    printf("pass\n");
+  return err;
+}
+
+static int fib_rec(int i)
+{
+  return (i>1) ? fib_rec(i-1) + fib_rec(i-2) : i;
+}
+
+static int fib(int i)
+{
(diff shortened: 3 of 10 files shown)


ganoam:
Hi people
Here is another simple optimization for synthesis on FPGAs. I have written a little custom program for a sanity check, since I don't think there is a test yet for checking the performance counters. It is pretty hacky so I don't know if you want me to commit it. Anyways, here is a patch in case somebody is interested. (appended .txt since github 'doesn't support .patch files')
[perfcnt-sanitycheck.patch.txt](https://github.com/openhwgroup/cv32e40p/files/5035889/perfcnt-sanitycheck.patch.txt)

to run: apply patch and `make perfcnt-vsim-run[-gui]` from folder `example_tb/core/`.

What it does: Configure the counters to count various events and combinations thereof while executing recursive fibonacci functions. Then check if the numbers make some sense. It is NOT an extensive test, just a proof of not being completely wrong.

ganoam:
Documentation in openhwgroup/core-v-docs#153.

Silabs-ArjanB:
@silabs-PaulZ Could you please have a look at @ganoam remark above that he maybe 'stumbled upon a related potential problem' and open a separate issue if you agree that it is a potential problem?

@ganoam Thank you very much for this pull request; I really appreciate the work you are doing here, but there are a couple of reasons why the changes do not match our design philosophy. I will add review comments in the code for that and I am also asking @davideschiavone to share his opinion.

Bender.yml
```diff
@@ -16,6 +16,7 @@ sources:
     - rtl/cv32e40p_alu_div.sv
     - rtl/cv32e40p_compressed_decoder.sv
     - rtl/cv32e40p_controller.sv
+    - rtl/cv32e40p_counter.sv
```
Silabs-ArjanB, line 19: Nice that you separated out the counter in a separate file

example_tb/core/Makefile
```diff
@@ -198,6 +198,34 @@ custom-vsim-run-gui: vsim-all custom/hello_world.hex
 custom-vsim-run-gui: ALL_VSIM_FLAGS += "+firmware=custom/hello_world.hex"
 custom-vsim-run-gui: vsim-run-gui
 
+#Perfcnt test
```
Silabs-ArjanB, line 201: I really appreciate you added a test case for this. The test case will have to move to the core-v-verif repos instead though.

example_tb/core/custom/perfcnt.c
```diff
@@ -0,0 +1,266 @@
+/*
```
Silabs-ArjanB, line 1: Test cases will be part of the core-v-verif repos instead.

rtl/cv32e40p_core.sv
```diff
@@ -30,11 +30,12 @@
...
-  parameter PULP_CLUSTER        =  0,                   // PULP Cluster interface (incl. p.elw)
-  parameter FPU                 =  0,                   // Floating Point Unit (interfaced via APU interface)
-  parameter PULP_ZFINX          =  0,                   // Float-in-General Purpose registers
-  parameter NUM_MHPMCOUNTERS    =  1
+  parameter PULP_XPULP         = 1,                   // PULP ISA Extension (incl. custom CSRs and hardware loop, excl. p.elw) !!! HARDWARE LOOP IS NOT OPERATIONAL YET !!!
+  parameter PULP_CLUSTER       = 0,                   // PULP Cluster interface (incl. p.elw)
+  parameter FPU                = 0,                   // Floating Point Unit (interfaced via APU interface)
+  parameter PULP_ZFINX         = 0,                   // Float-in-General Purpose registers
+  parameter NUM_MHPMCOUNTERS   = 1,
+  parameter WIDTH_MHPMCOUNTERS = 64
```
Silabs-ArjanB, line 38: For this core we were done adding parameters (as it implies additional verification space), so likely this will have to become a localparam.

rtl/cv32e40p_counter.sv
```diff
@@ -0,0 +1,82 @@
...
+    if (we) begin
+      counter_d = counter_load[CounterWidth-1:0];
+    end else if (counter_inc_i)begin
+      counter_d = counter_upd[CounterWidth-1:0];
+    end else begin
+      counter_d = counter[CounterWidth-1:0];
+    end
+  end
+
+`ifdef TARGET_XILINX
```
Silabs-ArjanB, line 45: We are aiming at a 100% elimination of ifdef and ifndef usage in RTL files (we will soon be there once we move the assertions into separate files), so we cannot accept new ifdefs popping up. In this case a solution could be to have two versions of this file, one for FPGA, one for ASIC (and solve the inclusion via the manifest file).

rtl/cv32e40p_counter.sv
```diff
@@ -0,0 +1,82 @@
...
+  // Set DSP pragma for supported Xilinx FPGAs
+  localparam int DspPragma = CounterWidth < 49  ? "yes" : "no";
+  (* use_dsp = DspPragma *) logic [CounterWidth-1:0] counter_q;
+
+  // DSP output register requires synchronous reset.
+  `define COUNTER_FLOP_RST posedge clk
```
Silabs-ArjanB, line 51: Same story for defines; we have been completely eliminating all define usage.

rtl/cv32e40p_counter.sv
```diff
@@ -0,0 +1,82 @@
...
+`else
+  logic [CounterWidth-1:0] counter_q;
+
+  `define COUNTER_FLOP_RST posedge clk or negedge rst_n
+`endif
+
+  // Counter flop
+  always_ff @(`COUNTER_FLOP_RST) begin
```
Silabs-ArjanB, line 59: This is the (only) serious issue as far as I am concerned: The ASIC version and the FPGA version will not be funtionally equivalent (for the scenario in which an async reset is asserted and then deasserted while the clock is not running). It is a must that the FPGA version is 100% functionally equivalent. I cannot judge if your area savings mostly came from this trick or not, so I don't know how bad it would be for the FPGA image to stick to the async reset.

rtl/cv32e40p_counter.sv
```diff
@@ -0,0 +1,82 @@
...
+    end else if (counter_inc_i)begin
+      counter_d = counter_upd[CounterWidth-1:0];
+    end else begin
+      counter_d = counter[CounterWidth-1:0];
+    end
+  end
+
+`ifdef TARGET_XILINX
+  // Set DSP pragma for supported Xilinx FPGAs
+  localparam int DspPragma = CounterWidth < 49  ? "yes" : "no";
```
Silabs-ArjanB, line 47: This ties the RTL file to one specific vendor; I am not a big fan of that, but as long as it doesn't break FGA synth for other vendors, then that would still be okay.

rtl/cv32e40p_counter.sv
```diff
@@ -0,0 +1,82 @@
+module cv32e40p_counter #(
+  parameter int CounterWidth = 32
```
Silabs-ArjanB, line 2: Parameters with all capitals in our RTL style

rtl/cv32e40p_cs_registers.sv
```diff
@@ -1403,124 +1409,144 @@ end //PULP_SECURE
...
+  if (NUM_MHPMCOUNTERS < 29) begin : g_mhpmcounter_non_implemented
+    logic [29-NUM_MHPMCOUNTERS-1:0] unused_mhpmcounter_we;
+    logic [29-NUM_MHPMCOUNTERS-1:0] unused_mhpmcounterh_we;
+    logic [29-NUM_MHPMCOUNTERS-1:0] unused_mhpmcounter_incr;
 
-            else if( (cnt_idx>2) && (cnt_idx<(NUM_MHPMCOUNTERS+3)))
-              // add +1 if any event is enabled and active
-              mhpmcounter_n[cnt_idx] = mhpmcounter_q[cnt_idx] +
-                                       |(hpm_events & mhpmevent_q[cnt_idx][NUM_HPM_EVENTS-1:0]) ;
+    // Lint tieoffs for unused bits
```
Silabs-ArjanB, line 1497: Nice

ganoam:
Hi @Silabs-ArjanB,

Thanks a lot for your feedback! I can understand your concerns about the reset. That might be a problem in certain cases. However, this change is essential for using the output pipeline registers of the DSP slices. If you decide it's not allowed to have sync reset, this optimization does not make sense for your core. 

The defines I don't like very much myself :) if you wish I will make separate files for asic and fpga.

I'll wait for further reviews and opinions before making changes...

Silabs-ArjanB:
Hi @ganoam Unfortunately I think that the 'logical equivalence' is more important than the 'FPGA area savings', so I do not see how we can overcome that issue without sacrificing the gains you achieved.

gautschimi:
mapping the performance counters to dsps as suggested allows to save quiet a few CLBs. It would be worth to include it. I think there is no strict need to reset the performance counter registers asynchronoulsy. they can easily be set in software before starting the counters.

the DSP attribute can probably also be set in the constraint file if you don't like it in the sourcecode. something like this should work:

set_property use_dsp yes [get_cells -hierarchical -filter { NAME =~ *cs_registers*/*counter_q* }]

Issue #439, opened 2020-08-06
Illegal Instruction Exception Raised Incorrectly - C.EBREAK


shetalani:
**RISC-V Specification:**

**Issue Description:**

Executing C.EBREAK instruction raises wrongly an illegal instruction exception. 

**Example:**

As shown below, the instruction 32'h00100073 (EBREAK) is decoded at time point t##0, where MSTATUS, MCAUSE and MEPC are updated at t##2 as if it had caused an illegal instruction exception.

![EBREAK_ill](https://user-images.githubusercontent.com/54367010/89549280-68ee3100-d808-11ea-8f4e-88648f556896.PNG)

**Top Level Parameters**
```
cv32e40p_core #(
    .FPU                      ( 0 ),
    .NUM_MHPMCOUNTERS         ( 1 ),
    .PULP_CLUSTER             ( 0 ),
    .PULP_XPULP               ( 0 ),
    .PULP_ZFINX               ( 0 )
) 
```
---
Product: OneSpin 360 DV-Verify
App: RVV
Tool's version: 2020.1.4

Silabs-ArjanB:
Hi @shetalani , could you check if this issue has been fixed now on the not yet merged https://github.com/Silabs-ArjanB/cv32e40p/tree/ArjanB_illegal_instr_exc branch. If not, can you show a new violating waveform (and I will remove the resolved label again).

Silabs-ArjanB:
![image](https://user-images.githubusercontent.com/40633348/91962861-65699f00-ed0d-11ea-80a2-5353b756495b.png)

Issue #449, opened 2020-08-18
Retracted interrupt can cause instructions to be skipped


Silabs-ArjanB:
In case an interrupt is made pending for only 1 cycle, then the controller will enter IRQ_FLUSH for one cycle and go back into DECODE the next cycle. If the instruction that is in decode experiences a stall at that point in time, then the instruction is skipped.

In the following example an interrupt is raised for 1 cycle just before the jr that will experience a couple of stall cycles. As you can see in the waveform the jr is not executed at all (the pc_id just increments with 4 all the time).

        csrw 0x300, 0x1f
        csrw 0x304, 0x1f
	addi x0, x0, 4
	addi x1, x0, 4
	addi x2, x0, 4
	addi x3, x0, 4
        lw x8, 0(x1)  # add additional wait state in data_rvalid_i to cause required stall
        jr 2034(x8)   # jr will stall because of x8 usage
	addi x4, x0, 1
	addi x5, x0, 1
	addi x6, x0, 1

![image](https://user-images.githubusercontent.com/40633348/90540672-d5125280-e181-11ea-96ba-236f223795b2.png)

The same problem can occur for other instructions stalling in ID when coinciding with a retracted interrupt. The issue is one of the underlying issues that causes #439-#442 as well.

davideschiavone:
Can we close this @Silabs-ArjanB ?

Issue #442, opened 2020-08-06
Illegal Instruction Exception Raised Incorrectly - FENCE


shetalani:
**RISC-V Specification:**

**Issue Description:**

Executing FENCE instruction raises wrongly an illegal instruction exception. 

**Example:**

As shown below, the instruction 32'h7b21008f (FENCE) is decoded at time point t##0, where MSTATUS, MCAUSE and MEPC are updated at t##2 as if it had caused an illegal instruction exception.

![FENCE_ill](https://user-images.githubusercontent.com/54367010/89547367-f54b2480-d805-11ea-8771-f64f01aebf56.PNG)

**Top Level Parameters**
```
cv32e40p_core #(
    .FPU                      ( 0 ),
    .NUM_MHPMCOUNTERS         ( 1 ),
    .PULP_CLUSTER             ( 0 ),
    .PULP_XPULP               ( 0 ),
    .PULP_ZFINX               ( 0 )
) 
```
---
Product: OneSpin 360 DV-Verify
App: RVV
Tool's version: 2020.1.4

Silabs-ArjanB:
Hi @shetalani , could you check if this issue has been fixed now on the not yet merged https://github.com/Silabs-ArjanB/cv32e40p/tree/ArjanB_illegal_instr_exc branch. If not, can you show a new violating waveform (and I will remove the resolved label again).

Silabs-ArjanB:
![image](https://user-images.githubusercontent.com/40633348/91963002-92b64d00-ed0d-11ea-985f-3519b7e0ded2.png)

Issue #441, opened 2020-08-06
Illegal Instruction Exception Raised Incorrectly - MRET


shetalani:
**RISC-V Specification:**

**Issue Description:**

Executing MRET instruction raises wrongly an illegal instruction exception. 

**Example:**

As shown below, the instruction 32'h30200073 (MRET) is decoded at time point t##0, where MSTATUS, MCAUSE and MEPC are updated at t##2 as if it had caused an illegal instruction exception.

![MRET_ill](https://user-images.githubusercontent.com/54367010/89549734-fdf12a00-d808-11ea-8ea2-7fcb65506529.PNG)

**Top Level Parameters**
```
cv32e40p_core #(
    .FPU                      ( 0 ),
    .NUM_MHPMCOUNTERS         ( 1 ),
    .PULP_CLUSTER             ( 0 ),
    .PULP_XPULP               ( 0 ),
    .PULP_ZFINX               ( 0 )
) 
```
---
Product: OneSpin 360 DV-Verify
App: RVV
Tool's version: 2020.1.4

Silabs-ArjanB:
Hi @shetalani , could you check if this issue has been fixed now on the not yet merged https://github.com/Silabs-ArjanB/cv32e40p/tree/ArjanB_illegal_instr_exc branch. If not, can you show a new violating waveform (and I will remove the resolved label again).

Silabs-ArjanB:
![image](https://user-images.githubusercontent.com/40633348/91962980-8af6a880-ed0d-11ea-9d35-c01869cc43c6.png)

PR #451, opened 2020-08-19
Fix for issues #438, #439, #440, #441, #442, #449

Silabs-ArjanB:
Stretched instr_valid_irq_flush_q pulse so that it is not missed if the instruction after a IRQ_FLUSH to DECODE transition experiences stalls (fix for #449 and partial fix for #439 #440 #441 #442) 

- [ ] Please review the stretch condition 'if (id_valid_i)'. The reason I picked that signal is because the 'tracer' judges the end of decoding by 'if ( id_valid && is_decoding )'


Silabs-ArjanB:
Commit 92b2300 takes care of setting illegal_insn_n back to 0 when a new instruction is in the DECODE state. Without this an instruction following an illegal instruction could get classified as an illegal instruction itself. See below waveform for an example. In this example 0x00045413 is an illegal instruction. However instruction 0x0000000f gets classified as illegal as well even though illegal_insn_i = 0 (because illegal_insn_n is simply a default copy of illegal_insn_q).

Together with the previous commit this should now fix #439 #440 #441 #442 

image

Silabs-ArjanB:
I want to do one more change in cv32e40p_decoder by replacing the following code:

              12'h302:  // mret
              begin
                illegal_insn_o = (PULP_SECURE) ? current_priv_lvl_i != PRIV_LVL_M : 1'b0;
                mret_insn_o    = ~illegal_insn_o;
                mret_dec_o     = 1'b1;
              end

              12'h002:  // uret
              begin
                illegal_insn_o = (PULP_SECURE) ? 1'b0 : 1'b1;
                uret_insn_o    = ~illegal_insn_o;
                uret_dec_o     = 1'b1;
              end

              12'h7b2:  // dret
              begin
                illegal_insn_o = !debug_mode_i;
                dret_insn_o    =  debug_mode_i;
                dret_dec_o     =  1'b1;
              end

by:

              12'h302:  // mret
              begin
                illegal_insn_o = (PULP_SECURE) ? current_priv_lvl_i != PRIV_LVL_M : 1'b0;
                mret_insn_o    = ~illegal_insn_o;
              end

              12'h002:  // uret
              begin
                illegal_insn_o = (PULP_SECURE) ? 1'b0 : 1'b1;
                uret_insn_o    = ~illegal_insn_o;
              end

              12'h7b2:  // dret
              begin
                illegal_insn_o = !debug_mode_i;
                dret_insn_o    =  debug_mode_i;
              end

and replacing all uses of mret_dec_o by mret_insn_o, uret_dec_o by uret_insn_o, and dret_dec_o by dret_insn_o.

To me it seems weird that these *_dec_o signals are 1 when their corresponding *_insn_o signal is 0.

Okay with above change 

- [ ] (then I will make a commit for that)?

Silabs-ArjanB:
- [ ] Hi @shetalani , could you please run OneSpin's formal checks on the proposed Silabs-ArjanB:ArjanB_illegal_instr_exc commit and check whether #438, #439, #440, #441, #442 have been solved or not (and if no new issues arose)? (Please mark the check box if okay and otherwise leave a comment on what is wrong)

shetalani:
> * [x]  Hi @shetalani , could you please run OneSpin's formal checks on the proposed Silabs-ArjanB:ArjanB_illegal_instr_exc commit and check whether #438, #439, #440, #441, #442 have been solved or not (and if no new issues arose)? (Please mark the check box if okay and otherwise leave a comment on what is wrong)

All the 5 issues are fixed. Cheers!

davideschiavone:
> 
> 
> Commit [92b2300](https://github.com/openhwgroup/cv32e40p/commit/92b2300524a3ff0d206dd3e8623d533115e602bb) takes care of setting illegal_insn_n back to 0 when a new instruction is in the DECODE state. Without this an instruction following an illegal instruction could get classified as an illegal instruction itself. See below waveform for an example. In this example 0x00045413 is an illegal instruction. However instruction 0x0000000f gets classified as illegal as well even though illegal_insn_i = 0 (because illegal_insn_n is simply a default copy of illegal_insn_q).
> 
> Together with the previous commit this should now fix #439 #440 #441 #442
> 
> Also added an assertion to check that mepc and illegal instruction encounter are consistent.
> 
> image

Hello @Silabs-ArjanB ,

in this example (just for my understanding) the following happens:

We have an illegal instruction
The core would like to go to FLUSH_EX but it is blocked by id_ready (maybe a stall in a previous ld/st instruction)
Then we receive an interrupt, I see the core going to IRQ_FLUSH
Then we go to TAKEN_ID
As now the illegal_insn_q is 1, the core thinks that the first instruction of the ISR is illegal.

Is that correct?

davideschiavone:
> illegal_insn_o



> mret_insn_o

I honestly don't know why we have mret_dec and not insn. It may be from old controlling logic but first impression looks ok. Can it happens that we have mret_insn_o switching from 1 to 0 or viceversa ?
I am almost sure it comes from PULP_SECURE race conditions.

rtl/cv32e40p_core.sv
```diff
@@ -1172,6 +1172,49 @@ module cv32e40p_core import cv32e40p_apu_core_pkg::*;
   // Assertions
   //----------------------------------------------------------------------------
 
+  // First illegal instruction decoded
+  logic         first_illegal_found;
+  logic [31:0]  expected_mepc;
+
+  always_ff @(posedge clk , negedge rst_ni)
```
davideschiavone: Can we have this part somehow explicitly saying that is not for synthesis but rather check? maybe we can move it to the BHV folder

davideschiavone (changes requested):

rtl/cv32e40p_core.sv
Silabs-ArjanB: All asssertions and related logic (like the one you commented on) inside `ifdef CV32E40P_ASSERT_ON `endif constructs will be moved out of the RTL files into the new sva (System Verilog Assertions' directory as proposed by Mike here: https://github.com/openhwgroup/cv32e40p/pull/421/files

Silabs-ArjanB:
> 
> 
> > Commit [92b2300](https://github.com/openhwgroup/cv32e40p/commit/92b2300524a3ff0d206dd3e8623d533115e602bb) takes care of setting illegal_insn_n back to 0 when a new instruction is in the DECODE state. Without this an instruction following an illegal instruction could get classified as an illegal instruction itself. See below waveform for an example. In this example 0x00045413 is an illegal instruction. However instruction 0x0000000f gets classified as illegal as well even though illegal_insn_i = 0 (because illegal_insn_n is simply a default copy of illegal_insn_q).
> > Together with the previous commit this should now fix #439 #440 #441 #442
> > Also added an assertion to check that mepc and illegal instruction encounter are consistent.
> > image
> 
> Hello @Silabs-ArjanB ,
> 
> in this example (just for my understanding) the following happens:
> 
> We have an illegal instruction
> The core would like to go to FLUSH_EX but it is blocked by id_ready (maybe a stall in a previous ld/st instruction)
> Then we receive an interrupt, I see the core going to IRQ_FLUSH
> Then we go to TAKEN_ID
> Then first instruction of the ISR is an instruction that needs FLUSH (fence in this example)
> As the illegal_insn_q is 1, the core thinks that the first instruction of the ISR is illegal.
> 
> Is that correct?

Hi @davideschiavone Yes, your understanding is correct. The fact that the first instruction in the ISR needs a flush is not really relevant for this example though. Any first instruction of the ISR would have been marked as illegal. As you can see in my fix I only forced illegal_insn_n back to 0 when we get back in the DECODE state (I did not try to alter the assignment to illegal_insn_q, as I do not fully oversee the use of that signal).

Silabs-ArjanB:
> 
> 
> > illegal_insn_o
> 
> > mret_insn_o
> 
> I honestly don't know why we have mret_dec and not insn. It may be from old controlling logic but first impression looks ok. Can it happens that we have mret_insn_o switching from 1 to 0 or viceversa ?
> I am almost sure it comes from PULP_SECURE race conditions.

I truly do not see the reason behind the current definition of the mret_dec_o, uret_dec_o, dret_dec_o signals. If I make them identical to mret_insn_o, uret_insn_o, dret_insn_o respectively, the CV32E40P is still logically equivalent. Anyway, I left that code as is (as it doesn't cause any issues either).

PR #473, opened 2020-09-02
add compressed prefix to trace log and waves of instructions

strichmo:

### DIFF
--- a/bhv/cv32e40p_tracer.sv
+++ b/bhv/cv32e40p_tracer.sv
@@ -117,6 +117,7 @@ module cv32e40p_tracer import cv32e40p_pkg::*;
     integer      cycles;
     logic [31:0] pc;
     logic [31:0] instr;
+    logic        compressed;
     string       str;
     reg_t        regs_read[$];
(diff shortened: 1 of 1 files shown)


Silabs-ArjanB (approved):

davideschiavone:
I understand the PR but the sintax in the tracer would be wrong and not in line to what the objdump would produce. 
In this way it is only adding **_c._** to an instruction that is already decompressed.

Silabs-ArjanB:
Hi @davideschiavone @strichmo Agreed with Davide, but I assume that Steve is aware of this as well and that he just wanted a way to annotate in the trace whether an instruction originated as a compressed instruction or not. (If we want to truly trace the compressed instructions, then the tracer needs to be updated to dump the instruction from before the predecoder at https://github.com/openhwgroup/cv32e40p/blob/c2f9ff51240bd07b33b9e1f58c11b107db1bba87/rtl/cv32e40p_if_stage.sv#L292 (which then needs to be properly timed). I assume that Steve is aware of this and did not need the original instruction code.

PR #475, opened 2020-09-03
:recycle: removed deprecated condition in DBG_TAKEN_ID

davideschiavone:
The DBG_TAKEN_ID is never used when single_step debug is active. This condition is from the time when the aligner was in the ID stage.


silabs-PaulZ (approved):
Looks good. We enter DBG_TAKEN_ID from DBG_FLUSH state only when debug_single_step is not asserted so this condition (in DBG_TAKEN_ID && debug_single_step) should never occur.

PR #476, opened 2020-09-03
:recycle: clean up code

davideschiavone:
Removed display condition in HWLoop and deprecated tb folder ending up in the master due to merge mistakes


Silabs-ArjanB (approved):
Looks good (the $warning, $stop will be replaced by UVM_ERROR in a next step, when the assertions are put into separate files in the sva directory)

Issue #481, opened 2020-09-04
Cause bits in dcsr should be read only


silabs-oysteink:
According to the riscv debug specification v0.13.2, the cause bits ([8:6]) in dcsr should be read only. During debug testing, I found that code executing in debug mode can actually write to these bits, causing a mismatch with the reference model.

eroom1966:
Documentation still shows this field as R/W - should this be recorded elsewhere ?

Silabs-ArjanB:
Hi @eroom1966 Could you please file a bug for that on https://github.com/openhwgroup/core-v-docs/issues?

silabs-oysteink:
Hi,

There is already an issue for the dcsr documentation, you could just update that one 😊
https://github.com/openhwgroup/core-v-docs/issues/163

Øystein

PR #483, opened 2020-09-04
Fix for #481

Silabs-ArjanB:

### DIFF
--- a/rtl/cv32e40p_cs_registers.sv
+++ b/rtl/cv32e40p_cs_registers.sv
@@ -1085,9 +1085,11 @@ end else begin //PULP_SECURE == 0
       end //csr_restore_mret_i
 
       csr_restore_dret_i: begin //DRET
-          // Restore to the recorded privilege level; if dcsr_q.prv is a non-supported mode, 
-          // then lowest privilege supported mode is selected (so always Machine Mode in this case).
-          priv_lvl_n = (dcsr_q.prv == PRIV_LVL_M) ? PRIV_LVL_M : PRIV_LVL_M;
+          // Restore to the recorded privilege level; if dcsr_q.prv is a non-supported mode,
+          // then the lowest privilege supported mode is selected. Therefore, as only Machine
+          // Mode is supported, priv_lvl_n will always be PRIV_LVL_M indepedent of the value
+          // of dcsr_q.prv.
+          priv_lvl_n = PRIV_LVL_M;
       end //csr_restore_dret_i
 
       default:;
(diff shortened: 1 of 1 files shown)


Silabs-ArjanB:
Hi @silabs-PaulZ , @silabs-oysteink This pull request should fix #481 Some of the reserved 0 bits were writeable as well and I made them always-read-as-0 instead. I would like to hear your opinion on 'dcsr_n.prv       = priv_lvl_q;  '; I think this bit should be fully R/W but as the spec says "If the encoding written is not supported or the debugger is not allowed to change to it, the hart may change to any supported privilege level." (So I think we should change that assignment into 'dcsr_n.prv = csr_wdata_int[1:0]; ' and then always simply ignore the written value (and basically pretend it is alway M mode).

silabs-oysteink:
I agree with your suggestion Arjan. The only side effect I can think of right now is if it affects the Imperas reference model, but if it does it's probably  a quick fix there as well.

silabs-PaulZ:
I was starting the process to update stepie to be read/writeable, but it would be fairly involved. I agree with tying it stepie=0. The implementation looks correct for tying this off (along with other WARL fields). I would like to look in more detail on the prv bitfield.

rtl/cv32e40p_cs_registers.sv
```diff
@@ -1073,9 +1085,9 @@ end else begin //PULP_SECURE == 0
       end //csr_restore_mret_i
 
       csr_restore_dret_i: begin //DRET
-        // restore to the recorded privilege level
-        // TODO: prevent illegal values, see riscv-debug p.44
-        priv_lvl_n = dcsr_q.prv;
+          // Restore to the recorded privilege level; if dcsr_q.prv is a non-supported mode, 
+          // then lowest privilege supported mode is selected (so always Machine Mode in this case).
+          priv_lvl_n = (dcsr_q.prv == PRIV_LVL_M) ? PRIV_LVL_M : PRIV_LVL_M;
```
silabs-PaulZ, line 1090: I would consider priv_lvl_n = PRIV_LVL_M in this case for any code coverage issues. Hopefully we can refactor this file soon to reduce the duplication issues.

silabs-PaulZ (approved):

rtl/cv32e40p_cs_registers.sv
MikeOpenHWGroup: > Hopefully we can refactor this file soon to reduce the duplication issues.

Its getting late in the day for RTL refactoring exercises.  If you are serious about this, please open an issue so the cost/risk/benefit can be discussed.

Silabs-ArjanB: > 
> 
> I would consider priv_lvl_n = PRIV_LVL_M in this case for any code coverage issues. Hopefully we can refactor this file soon to reduce the duplication issues.

Okay, I will rewrite like that and merge this pull request. Thanks.

PR #487, opened 2020-09-05
add a insn_compressed signal in the tracer that align with current retired instruction

strichmo:
as hint to the functional coverage that the instruction was originally compressed

This is fastest route to fixing issues with current RV32ISA functional coverage model.  

Long term we should make some structural changes to the tracer such that it properly reflects the ISA-level instructions entering the pipeline.  I will open a separate issue to track that.


Silabs-ArjanB (approved):
Agreed with the short and long term solution

Silabs-ArjanB:
@strichmo Just let me know what you want the final (longer term) tracer format to look like, then we will implement that

PR #469, opened 2020-09-01
Prevented IRQ related CSR update to be overwritten by duplicate regul…

Silabs-ArjanB:
Prevented IRQ related CSR update to be overwritten by duplicate regular CSR write. CSR writes are now only performed during the initial cycle in EX.

Fix for CSR + IRQ conflict from #460 

Signed-off-by: Arjan Bink <Arjan.Bink@silabs.com>

Issue #460, opened 2020-08-26
interrupt CSR checks fail when interrupt occurs during div instruction


strichmo:
When enabling random interrupts with the ISS in the core-v-verif testbench CSR miscompares can occur on instruction retirement boundaries.

The testcase is an interrupt that is acknowledge during a div instruction.  See the first figure below.

![iss_irq0](https://user-images.githubusercontent.com/56401023/91345174-947b9000-e7a4-11ea-9861-d77243cb9a52.png)

In this case the interrupt is acknowledge 2 instruction retirements before the machine mode interrupt vector is entered.  Therefore the PC is miscompared against the ISS.

To address this I added a staging of the "deferint" signal which masks the irq assertion in the ISS to the second instruction.  This enables the PCs to compare properly between RTL and the ISS.

![iss_irq1](https://user-images.githubusercontent.com/56401023/91345354-e0c6d000-e7a4-11ea-9540-20b2133349ae.png)

However with this case, the CSRs for machine mode interrupts (mepc, mstatus, mcause) will miscompare on the first instruction retirement because the RTL updates the CSRs immediately upon interrupt acknowledgement.

### Steps to Reproduce
To replicate the first condition (PC miscompare):

1. Use this branch and fork:
https://github.com/strichmo/core-v-verif/tree/strichmo/irq_random_fail_0

2. Run this in cv32/sim/uvmt_cv32
% make corev-dv gen_corev_rand_interrupt_test NUM_TESTS=1 SEED=1
% make custom CUSTOM_PROG=corev_rand_interrupt_test_0 USER_RUN_FLAGS="+gen_irq_noise +UVM_MAX_QUIT_COUNT=20,NO" SIMULATOR=xrun

To replicate with the proposed "deferint_prime" fixes:

1. Use this branch and fork:
https://github.com/strichmo/core-v-verif/tree/strichmo/irq_random_fail_fix_0

2. Run the same tests:
% make corev-dv gen_corev_rand_interrupt_test NUM_TESTS=1 SEED=1
% make custom CUSTOM_PROG=corev_rand_interrupt_test_0 USER_RUN_FLAGS="+gen_irq_noise +UVM_MAX_QUIT_COUNT=20,NO" SIMULATOR=xrun

UVM_ERROR @ 72790.000 ns : uvmt_cv32_step_compare.sv(79) reporter [Step-and-Compare] mcause expected=0x80000003 and actual=0x8000001d PC=0x00003244
[...]

PR #491, opened 2020-09-08
:bug: fix csr_status for MEPC

davideschiavone:
This PR replaces the #469 for the issue found in #460


Silabs-ArjanB (changes requested):
The fix has only be applied for MEPC. I expect that the same bug can occur for CSR_MCAUSE which is also a R/W CSR that is affected by a taken interrupt. Maybe even the read-only CSR_PRIVLV should be included in this fix as it can be read by a CSR instruction and such a read should not reflect the status of an interrupt on a following instruction (of course in CV32E40P this CSR will always return machine mode, so for now this doesn't matter; I expect that a CSR_PRIVLV read could go wrong in an implementation including User mode(

Silabs-ArjanB (approved):

Silabs-ArjanB:
Hi @davideschiavone Thanks; good to merge as far as I am concerned!

Issue #485, opened 2020-09-04
Unique case violations at t=0


MikeOpenHWGroup:
When running a simulation on the 6fbd88c645d2b51c316af6eda79bab3e4c284093 hash of the RTL, **XRUN** reports the following unique case violations at t=0:

```
xmsim: *W,NOCOND: Unique case violation:  Every case item expression was false.
            File: /home/v51366/mydata/GitHubRepos/openhwgroup/core-v-verif/master/core-v-cores/cv32e40p/rtl/cv32e40p_cs_registers.sv, line = 1094, pos = 14
           Scope: uvmt_cv32_tb.dut_wrap.cv32e40p_wrapper_i.core_i.cs_registers_i
            Time: 0 FS + 72

xmsim: *W,NOCOND: Unique case violation:  Every case item expression was false.
            File: /home/v51366/mydata/GitHubRepos/openhwgroup/core-v-verif/master/core-v-cores/cv32e40p/rtl/cv32e40p_prefetch_controller.sv, line = 168, pos = 14
           Scope: uvmt_cv32_tb.dut_wrap.cv32e40p_wrapper_i.core_i.if_stage_i.prefetch_buffer_i.prefetch_controller_i
            Time: 0 FS + 72

xmsim: *W,NOCOND: Unique case violation:  Every case item expression was false.
            File: /home/v51366/mydata/GitHubRepos/openhwgroup/core-v-verif/master/core-v-cores/cv32e40p/rtl/cv32e40p_obi_interface.sv, line = 131, pos = 16
           Scope: uvmt_cv32_tb.dut_wrap.cv32e40p_wrapper_i.core_i.if_stage_i.prefetch_buffer_i.instruction_obi_i.genblk1
            Time: 0 FS + 72
```
Note that there are no other errors in the simulation.

The same violations are reported by **DSIM**.   I have confirmed that the violations go away if the `unique case` is replaced with `unique0 case`.

### Steps to Reproduce
```
$ git clone https://github.com/openhwgroup/core-v-verif
$ cd core-v-verif/cv32/sim/uvmt_cv32
$ make SIMULATOR=xrun sanity
```

Silabs-ArjanB:
This looks like an error of core-v-verif itself. Specifically not all inputs to cv32e40p_core are assigned at time 0 and some are not driven at at any time. All input signals need to be valid at all times, especially the clock input (which is not the case).

[...]

PR #492, opened 2020-09-08
Replaced unique case by case to prevent simulation warnings

Silabs-ArjanB:


Silabs-ArjanB:
This fixes https://github.com/openhwgroup/cv32e40p/issues/485

Issue #429, opened 2020-07-22
HPM counter logic 'possibly' not implemented correctly


silabs-PaulZ:
@davideschiavone ,

This issue was found by inspecting the RTL. It also serves as a reminder to write a test for HPM. The RTL looks to increment the jump, branch, and branch_taken based on a signal asserted during the decode phase (e.g. jump_i) along with a pipelined control signal (id_valid_q). 

https://github.com/openhwgroup/cv32e40p/blob/1607d8b675864db1aa013364fd4444a665331830/rtl/cv32e40p_cs_registers.sv#L1352-L1363

Perhaps this should be replaced with the non-pipelined instruction retired condition (e.g. jump_i & id_valid_i & is_decoding_i).
hash: 8a626b912ba5491c9fb384a4e498554ca579cae3

Silabs-ArjanB:
I can confirm (at least) two bugs in above code:

1) The performance counter registers software writes do not take effect on the correct instruction boundaries (the fix for this is the the insertion of flush states as is done for many other CSRs)
2) The id_valid_q signal is not necessarily aligned with jump_i as is shown in below waveform (this can be fixed with qualifying jump_i differently). (Maybe similar bugs exist for the other id_valid_q uses)

  45026ns            4498 00000788 00000013 nop                                 
  45036ns            4499 0000078c 00000013 nop                                 
  45046ns            4500 00000790 00012203 lw               x4, 0(x2)            x4=xxxxxxxx  x2:0000f9f0  PA:0000f9f0
  45066ns            4502 00000794 2cc00267 jalr             x4, x0, 716          x4=00000798
  45076ns            4503 000002cc 03010113 addi             x2, x2, 48           x2=0000fa20  x2:0000f9f0

![image](https://user-images.githubusercontent.com/40633348/92947508-62cb2000-f458-11ea-9d8d-b1ae7a0b8a6a.png)

Silabs-ArjanB:
Formal shows that id_valid_q is implied by branch_i, so following will be simplified by removing id_valid_q:

assign hpm_events[8]  = branch_i                   & id_valid_q;       // nr of branches (conditional) 
assign hpm_events[9]  = branch_i & branch_taken_i  & id_valid_q;       // nr of taken branches (conditional) 

Following will be written more intuitively:

 assign hpm_events[3]  = jr_stall_i & id_valid_q;                       // nr of jump register hazards 

[...]

PR #504, opened 2020-09-15
Fix for #429

Silabs-ArjanB:


Silabs-ArjanB:
@silabs-PaulZ You can find the C code I used when doing this fix here: https://github.com/openhwgroup/core-v-verif/issues/239

silabs-PaulZ:
Hi @Silabs-ArjanB , Made some minor updates to the C-code to make it run for me.

[hpmcounter.txt](https://github.com/openhwgroup/cv32e40p/files/5258920/hpmcounter.txt)

silabs-PaulZ (commented):
Most of this seems correct, except for hpm_events[2] (load_stall).  If you get a IRQ or DEBUG flush, you can get +2 for the stall as opposed to the intended +1. 

Add the following assertion in the cv32e40p_cs_registers.sv:

```
  a_count_load_stall_once : assert property
  (
   @(posedge clk)  disable iff (!rst_n)
   (hpm_events[2]   |=> !(hpm_events[2]) throughout inst_ret [->1])
   );
```
The event gets counted before the IRQ or DEBUG flush and then after then again after the flush.

image

On a side note, it seems a little odd that flush will not wait for OBI GNT and write back to finish before proceeding.

I cannot comment on the APU counter events.

Silabs-ArjanB:
@davideschiavone Could you please comment about Paul's question on the DBG_FLUSH not actually performing a pipeline flush, what is the intent of that state? (At first glance this might also be related to https://github.com/openhwgroup/cv32e40p/issues/509)

Silabs-ArjanB:
>> Most of this seems correct, except for hpm_events[2] (load_stall). If you get a IRQ or DEBUG flush, you 
>> can get +2 for the stall as opposed to the intended +1.

I made a fix that load and jump hazards are not counted for flushed instructions. This however does not solve the +2 increments for IRQ and Debug, but I believe the behavior is correct like this. The instruction after an IRQ or Debug flush is a new instruction which should be counted with respect to hazards as well (even if this new instruction experiences a hazard with a slow load that is still in the pipeline after a debug flush).

Issue #430, opened 2020-07-22
Fix TODOs in RTL


Silabs-ArjanB:
The following 'todo's in the RTL should be addressed as they reflect 'potential' bugs:

- cv32e40p_cs_registers.sv (https://github.com/openhwgroup/cv32e40p/blob/1607d8b675864db1aa013364fd4444a665331830/rtl/cv32e40p_cs_registers.sv#L1023):        // TODO: prevent illegal values, see riscv-debug p.44
- cv32e40p_controller.sv:                    // TODO: handle ebrk_force_debug_mode plus single stepping over ebreak
- cv32e40p_controller.sv:                            ctrl_fsm_ns = FLUSH_EX; // TODO: flush ex
- cv32e40p_controller.sv:                  // TODO: why is this here, signal only needed for async exceptions
- cv32e40p_controller.sv:              //TODO: is illegal when not in debug mode

The others are enhancements or issues that are not actual for CV32E40P:

- cv32e40p_id_stage.sv:  // TODO: check if this can be shared with the bit-manipulation unit
- cv32e40p_id_stage.sv:  // TODO: check if sign-extension stuff works well here, maybe able to save
- cv32e40p_mult.sv:  assign int_is_msu = (operator_i == MUL_MSU32); // TODO: think about using a separate signal here, could prevent some switching
- cv32e40p_decoder.sv:    // TODO: this section should eventually be moved out of the decoder

- cv32e40p_cs_registers.sv:              // TODO: correctly handled?
- cv32e40p_cs_registers.sv:                // TODO: correctly handled?
- cv32e40p_cs_registers.sv:                // TODO: correctly handled?
- cv32e40p_cs_registers.sv:          // TODO: prevent illegal values, see riscv-debug p.44

PR #516, opened 2020-09-22
Fix Todo: remove comment flush ex

silabs-PaulZ:
Fixes two TODOs in the cv32e40p_controller.sv RTL based on issue #430

The comment is not clear. The code is written correctly whereby an illegal instruction or ecall instruction will enter the FLUSH_EX state while running in single-step mode.

@bluewww , Can you please review this PR as you may have entered this TODO comment on a previous change:
https://github.com/openhwgroup/cv32e40p/commit/efc02fc77cf250d3c2c618ae406ca6cbe0689e7e#diff-f8105fd1c66cefe40c50f701d9f3108eR565

### DIFF
--- a/rtl/cv32e40p_controller.sv
+++ b/rtl/cv32e40p_controller.sv
@@ -646,7 +646,7 @@ module cv32e40p_controller import cv32e40p_pkg::*;
 
                         illegal_insn_i | ecall_insn_i:
                         begin
-                            ctrl_fsm_ns = FLUSH_EX; // TODO: flush ex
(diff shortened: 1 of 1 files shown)


davideschiavone (approved):

davideschiavone:
The TODO is probably refered to "TODO: verify that this is correct". Merged. In case @bluewww thinks it's wrong we should open a new issue

bluewww:
Looks like spurious TODO's to me. I don't remember what I was thinking back then.

PR #518, opened 2020-09-22
Fix Todo: remove TODO comment to re-locate code

silabs-PaulZ:
Fixes TODO in the cv32e40p_decoder.sv RTL based on issue #430

Todo comment suggests moving a piece of code out of decoder. Code has been there for 5 years and seems relevant to alu operation based on data_misalignment.

@atraber , Can you please review and see if the TODO comment is OK to be removed. The comment was added with this change:
https://github.com/openhwgroup/cv32e40p/commit/9858caff47a4213d0222ee19290db88facd6be6a

### DIFF
--- a/rtl/cv32e40p_decoder.sv
+++ b/rtl/cv32e40p_decoder.sv
@@ -2886,7 +2886,6 @@ module cv32e40p_decoder import cv32e40p_pkg::*; import cv32e40p_apu_core_pkg::*;
     end
 
     // misaligned access was detected by the LSU
-    // TODO: this section should eventually be moved out of the decoder
     if (data_misaligned_i == 1'b1)
     begin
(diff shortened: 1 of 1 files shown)


davideschiavone:
If I may get an interpretation of @atraber  comment, I guess he means that the decoder is usually to generate signals starting from the instruction to (in fact) decode. Whereas in this case, the logic coming from the EX stage (i.e. data_misaligned_i) is exploiting the same piece of RTL of the decoder, instead of actually muxing those signal in the ID_stage. I make a PR

davideschiavone:
If #524 gets accepted, we can close this PR without merging it

Silabs-ArjanB:
Rejected in favor of https://github.com/openhwgroup/cv32e40p/pull/524

Issue #484, opened 2020-09-04
cv32e40p_register_file_test_wrap has unused BIST ports


MikeOpenHWGroup:
Module `cv32e40p_register_file_test_wrap` has a set of un-used BIST ports.  It is instantiated in module `cv32e40p_id_stage`:

```
cv32e40p_register_file_test_wrap
  #(
    .ADDR_WIDTH(6),
    .FPU(FPU),
    .PULP_ZFINX(PULP_ZFINX)
  )
  registers_i
  (
    .clk                ( clk                ),
    .rst_n              ( rst_n              ),

// sniped port-map....

     // BIST ENABLE
     .BIST        ( 1'b0                  ), // PLEASE CONNECT ME;

     // BIST ports
     .CSN_T       ( 1'b0                ), // PLEASE CONNECT ME; Synthesis will remove me if unconnected
     .WEN_T       ( 1'b0                ), // PLEASE CONNECT ME; Synthesis will remove me if unconnected
     .A_T         ( 6'b0                ), // PLEASE CONNECT ME; Synthesis will remove me if unconnected
     .D_T         (32'b0                ), // PLEASE CONNECT ME; Synthesis will remove me if unconnected
     .Q_T         (                     )
  );
```

As far as I am aware, this is the only BIST interface in the core and the test_wrap is the only module that provides (or needs) BIST access.  Also, BIST is not mentioned in the User Manual.   I recommend that we either:

1. Remove the BIST logic and ports altogether, or,
2. Bring these ports to the top-level of the core and document their use in the User Manual.

This issue was found by looking at holes in [code coverage](https://openhwgroup.metrics.ca/cv32e40p_verif_3-core-v-verif/coverage/code/regressionRuns/cc3c865f-5cee-4faa-a9de-dff8333f55c0) from a Compliance test-suite regression run on the 6fbd88c645d2b51c316af6eda79bab3e4c284093 hash of this repo.

Silabs-ArjanB:
Hi @davideschiavone , what do you think about this. For me option 2 is not really an option. Either we leave it like this or (preferred by us) we remove this BIST related logic all together (it is not needed as far as I am concerned)

davideschiavone:
Let me ask the guy that made that change in the RF. Coming back soon

Silabs-ArjanB:
[...]

PR #521, opened 2020-09-23
Removed register file bist wrapper #484

Silabs-ArjanB:


Silabs-ArjanB:
Fix for #484

PR #525, opened 2020-09-24
:bug: fix #372

### LINKED ISSUE #372: pv.cplxconj decoding and tracing
pv.cplxconj is documented correctedly. However based on the RTL comments and tracer outputs it seems as if pv.cplxconj.b and pv.cplxconj.h exist.

- The tracer should be adapted to prevent adding the .b/.h suffix for the pv.cplxconj  instruction.
- The RTL (cv32e40p_decoder.sv) should be adapted to refer to pv.cplxconj instead of pv.cplxconj.h in its comments
- The RTL (cv32e40p_decoder.sv) should be adapted to cause an illegal instruction exception for a 'pv.cplxconj instruction with bit 12 set to 1).

Related issues: https://github.com/openhwgroup/core-v-docs/issues/95 and #363

davideschiavone:
PArt of issue #372 was solved in PR #479 . Here I removed the .h extention

### DIFF
--- a/bhv/cv32e40p_tracer.sv
+++ b/bhv/cv32e40p_tracer.sv
@@ -695,6 +695,7 @@ module cv32e40p_tracer import cv32e40p_pkg::*;
                2'b11: mnemonic = instr[25] == 1'b0 ? "pv.cplxmul.r.div8" : "pv.cplxmul.i.div8";
             endcase
             str_sci = "";
+            str_hb  = "";
           end
 
           6'b011011: begin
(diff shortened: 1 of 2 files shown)


davideschiavone:
Before mergining it we should add the illegal condition in the .b case --> see #372

Silabs-ArjanB (approved):

davideschiavone:
I added the illegal condition. Can you please check? @Silabs-ArjanB

Silabs-ArjanB (commented):
The original issue was specific to pv.cplxconj. I see you applied the same fix for the other complex instructions as well. It looks consistent, but I cannot comment on what the intention was for all those instructions.

davideschiavone:
It's all the same, they all act on complex numbers, so a vector of 2 16b operands packed in a 32b register. I modified the documentation as well.

Silabs-ArjanB (approved):

PR #527, opened 2020-09-24
Removed mult related todo. mentioned signal comes from flop so impact…

Silabs-ArjanB:
… will be small if any)


Silabs-ArjanB:
https://github.com/openhwgroup/cv32e40p/issues/430

PR #528, opened 2020-09-30
Cleaned up RTL by only supporting SHARED_FPU == 2

Silabs-ArjanB:
Cleaned up RTL under the assumption that we only want to support SHARED_FP = 2.

Under that assumption (i.e. after changing SHARED_FP from 0 to 2 in the 'golden RTL' the RTL from this pull request and the 'golden RTL' are equivalent for the following two configurations:

Config 1:

  parameter PULP_XPULP          =  1
  parameter PULP_CLUSTER        =  1
  parameter FPU                 =  1
  parameter PULP_ZFINX          =  1
  parameter NUM_MHPMCOUNTERS    =  1

Config 2:

  parameter PULP_XPULP          =  0
  parameter PULP_CLUSTER        =  0
  parameter FPU                 =  0
  parameter PULP_ZFINX          =  0
  parameter NUM_MHPMCOUNTERS    =  1

When using the current 'master' RTL as golden, which has SHARED_FP = 0, the pull request is equivalent for the following configurations:

Config 3:

  parameter PULP_XPULP          =  0
  parameter PULP_CLUSTER        =  0
  parameter FPU                 =  0
  parameter PULP_ZFINX          =  0
  parameter NUM_MHPMCOUNTERS    =  1

Config 4:

  parameter PULP_XPULP          =  1
  parameter PULP_CLUSTER        =  1
  parameter FPU                 =  0
  parameter PULP_ZFINX          =  0
  parameter NUM_MHPMCOUNTERS    =  1

(In this case setting FPU = 1 will lead to a non equivalent circuit as we would be comparing RTL with SHARED_FP == 0 against RTL with SHARED_FP == 2).

This pull request do not include the ALU simplifications that should be made as well.


Silabs-ArjanB:
The second commit is a further cleanup under the assumption that we only ever want to support:

SHARED_FP_DIVSQRT == 1
FP_DIVSQRT == FPU

Under that assumption (after changing SHARED_FP_DIVSQRT from 0 to 1 in the golden RTL) the following configurations are equivalent to the RTL of the pull request:

parameter PULP_XPULP          =  0
parameter PULP_CLUSTER        =  0
parameter FPU                 =  0
parameter PULP_ZFINX          =  0
parameter NUM_MHPMCOUNTERS    =  1

and

parameter PULP_XPULP          =  1
parameter PULP_CLUSTER        =  1
parameter FPU                 =  1
parameter PULP_ZFINX          =  1
parameter NUM_MHPMCOUNTERS    =  1

MikeOpenHWGroup:
> Cleaned up RTL under the assumption that we only want to support SHARED_FP = 2.

Sorry, I may have missed something.  Where did the requirement for this originate?   Is it captured somewhere?

rtl/cv32e40p_decoder.sv
```diff
@@ -166,20 +165,18 @@ module cv32e40p_decoder import cv32e40p_pkg::*; import cv32e40p_apu_core_pkg::*;
   localparam APUTYPE_DSP_MULT   = (SHARED_DSP_MULT)       ? 0 : 0;
   localparam APUTYPE_INT_MULT   = (SHARED_INT_MULT)       ? SHARED_DSP_MULT : 0;
   localparam APUTYPE_INT_DIV    = (SHARED_INT_DIV)        ? SHARED_DSP_MULT + SHARED_INT_MULT : 0;
-  localparam APUTYPE_FP         = (SHARED_FP)             ? SHARED_DSP_MULT + SHARED_INT_MULT + SHARED_INT_DIV : 0;
-  localparam APUTYPE_ADDSUB     = (SHARED_FP)             ? ((SHARED_FP==1) ? APUTYPE_FP   : APUTYPE_FP)   : 0;
-  localparam APUTYPE_MULT       = (SHARED_FP)             ? ((SHARED_FP==1) ? APUTYPE_FP+1 : APUTYPE_FP)   : 0;
-  localparam APUTYPE_CAST       = (SHARED_FP)             ? ((SHARED_FP==1) ? APUTYPE_FP+2 : APUTYPE_FP)   : 0;
-  localparam APUTYPE_MAC        = (SHARED_FP)             ? ((SHARED_FP==1) ? APUTYPE_FP+3 : APUTYPE_FP)   : 0;
+  localparam APUTYPE_FP         = SHARED_DSP_MULT + SHARED_INT_MULT + SHARED_INT_DIV;
```
davideschiavone, line 166: In my opinion we should also remove SHARED_DSP_MULT and SHARED_INT_MULT and SHARED_INT_DIV

davideschiavone (commented):
I would also like to have the expert person's opinion @stmach

davideschiavone:
Hi @MikeOpenHWGroup, this is just code clean up :) no new specs nothing of nothing. Just removing dead code

rtl/cv32e40p_decoder.sv
Silabs-ArjanB: @davideschiavone Can you briefly explain the original idea behind for example SHARED_DSP_MULT? Did it enable sharing of part of the current ALU with other cores by putting it external to the core, and if so, do you now see that as a feature that will not be needed in the future anymore?

davideschiavone: It was an idea similar to have an accelerator outside doing some SIMD or DSP operation (like now the FPU)
It good as feature for CV32E40 (no p) as we want to do this anyway, but for the P version let's jut removed old, obsolete, not maintained code

Silabs-ArjanB: Sounds good; let's wait for @stmach 's comment then before proceeding.

stmach: LGTM!
While these parameters used to mainly affect the configuration of an external shared unit (for multi-core scenarios), they were necessary inside the core only for 'addressing' which shared unit shall receive the instruction (which had been made dependent on the number of different units instead of a fixed enumeration).
As AFAIK you're planning on a proper 'accelerator interface' anyways, there's no use in keeping these.

Silabs-ArjanB: Hi @davideschiavone , so I can do another commit to simplify the RTL assuming the following will localparams will remain fixed as follows?

SHARED_DSP_MULT     =  0
SHARED_INT_MULT     =  0
SHARED_INT_DIV      =  0

davideschiavone: yes

davideschiavone (approved):

PR #532, opened 2020-10-01
FPU now attached at tb level with APU interface

davideschiavone:


example_tb/core/cv32e40p_tb_subsystem.sv
```diff
@@ -99,16 +117,16 @@ module cv32e40p_tb_subsystem
...
-         .apu_master_gnt_i       (1'b0                   ),
-         .apu_master_operands_o  (                       ),
-         .apu_master_op_o        (                       ),
-         .apu_master_type_o      (                       ),
-         .apu_master_flags_o     (                       ),
-         .apu_master_valid_i     (1'b0                   ),
-         .apu_master_result_i    (32'b0                  ),
-         .apu_master_flags_i     (5'b0                   ),
+         .apu_master_req_o       ( apu_req              ),
+         .apu_master_ready_o     (                      ),
```
Silabs-ArjanB, line 121: Does the fact that you are not connecting apu_master_ready_o and apu_master_type_o mean that these signals will never be used (and can be removed from the CV32E40P) or are there other scenarios in which these are needed?

Silabs-ArjanB (approved):

davideschiavone:
This one I merged, if we  can change the interface, then I remove in another PR both apu_master_type_o and apu_master_ready_o. We should also change the name by removing master as it is not consistent with the others interfaces. (i.e. data and instr mem interfaces)

Silabs-ArjanB:
Hi @davideschiavone Okay, let's do the pin removal and renaming after RTL freeze as it is too disruptive right now.

Issue #529, opened 2020-09-30
Wrong dcsr.cause after sleep


silabs-oysteink:
## Wrong dcsr.cause after sleep
Cause bits in dcsr seems to wrong if the first instruction decoded after WFI + debug_req_i is an EBREAK.

1. Core is sleeping
2. debug_req_i is asserted
3. Core wakes up, fetches an ebreak instruction
4. Core enters debug mode, sets cause to 0x1 (ebreak)
5. Ebreak instruction does not retire

<img width="605" alt="ebreak_2" src="https://user-images.githubusercontent.com/66771756/94662635-f7fb5f00-0308-11eb-8d10-58a897c8e485.png">

Debug spec v 0.13.2 states the following:

> If halt is requested while wfi is executing, then the hart must leave the stalled state, completing
this instruction’s execution, and then enter Debug Mode.

In this case, I would expect the core to disregard the ebreak and enter debug mode with cause=0x3 (external debugger). Instead it enters debug as if the ebreak caused it, although the ebreak does not retire.

### Component
**Component:RTL**

### Steps to Reproduce
1. Clone my fork of core-v-verif:  https://github.com/silabs-oysteink/core-v-verif.git
2. Check out branch silabs-oysteink_random-debug
3. from cv32/sim/uvmt_cv32:
4: make comp_corev-dv gen_corev-dv TEST=corev_rand_debug START_INDEX=0 NUM_TESTS=1
5: make test TEST=corev_rand_debug RUN_INDEX=0 WAVES=YES USE_ISS=NO

I've disabled the ISS as the current state of the repo requires an invalid license. Assertion will fire regardlessly.

Silabs-ArjanB:
@silabs-PaulZ I expect that exactly the same issue will happen when we try to assert debug_req_i upon cold start while the first regular program instruction would have been an ebreak.

silabs-PaulZ:
Thank you @silabs-oysteink for finding this. 

I submitted a fix with PR #531. This corner case occurs with WFI, reset, and also in general code execution.

Below are waveforms showing the new register ebrk_debug_entry_q that will capture if entry is due to a ebreak instruction. The top half of the waveform shows the scenario where debug_req preempts the ebreak instruction and cause is set to `haltreq`. The bottom half is where debug_req is after the taken ebreak instruction, asserting ebrk_debug_entry_q, and setting cause to `ebreak`.

[...]

PR #538, opened 2020-10-02
fix priority of debug_req and single step

silabs-PaulZ:
Replaced the debug_ebrk_entry register add in #531  with debug_req_entry register. This will assert when debug entry is caused by debug_req at the time it preempts any instruction being decoded. This will fix the issues reported in #529 

### DIFF
--- a/rtl/cv32e40p_controller.sv
+++ b/rtl/cv32e40p_controller.sv
@@ -214,7 +214,7 @@ module cv32e40p_controller import cv32e40p_pkg::*;
   logic ebrk_force_debug_mode;
   logic is_hwlp_illegal, is_hwlp_body;
   logic illegal_insn_q, illegal_insn_n;
-  logic ebrk_debug_entry_q, ebrk_debug_entry_n;
+  logic debug_req_entry_q, debug_req_entry_n;
 
   logic hwlp_end0_eq_pc;
   logic hwlp_end1_eq_pc;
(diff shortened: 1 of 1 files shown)


rtl/cv32e40p_controller.sv
```diff
@@ -1196,26 +1198,18 @@ module cv32e40p_controller import cv32e40p_pkg::*;
...
+          if(debug_mode_q                          |
+             trigger_match_i                       |
+             (ebrk_force_debug_mode & ebrk_insn_i) |
+             data_load_event_i                     |
+             debug_req_entry_q                     )
+            begin
+              ctrl_fsm_ns = DBG_TAKEN_ID;
+            end else if (debug_single_step_i) begin
+              ctrl_fsm_ns = DBG_TAKEN_IF;
+            end
```
Silabs-ArjanB, line 1212: Should at least one of the conditions above be true when we arrive in DBG_FLUSH? Now the code would seemingly allow a DBG_FLUSH -> DBG_FLUSH state transition, can that really happen; if not let's add an assertion and maybe remove the 'if (debug_single_step_i)

Silabs-ArjanB (approved):

rtl/cv32e40p_controller.sv
silabs-PaulZ: Agreed, I added an assertion in the PR #541 to ensure FSM is not stuck in DBG_FLUSH state.

PR #539, opened 2020-10-02
Riviera-PRO support

dawidzim:
Hey!
This is a patch which added support for Riviera-PRO simulator.

### DIFF
--- a/example_tb/core/Makefile
+++ b/example_tb/core/Makefile
@@ -44,6 +44,21 @@ VSIM_DEBUG_FLAGS	= -debugdb
 VSIM_GUI_FLAGS          = -gui -debugdb
 VSIM_SCRIPT             = vsim.tcl
 
+# rivierapro configuration
+ALIB                    = "vlib"
+AWORK                   = work
+
+ALOG                    = vlog
+ALOG_FLAGS              = # user defined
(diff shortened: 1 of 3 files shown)


Silabs-ArjanB (commented):
No ifdefs are allowed in the RTL. Cannot accept this. What is 'wrong' with the FIFO RTL?

dawidzim:
Hey @Silabs-ArjanB 
I remove `ifdef`. What you think now about this?

Silabs-ArjanB:
Hi @dawidzim You also removed the 'unique' keyword. Can you explain what is wrong with the original code please?

dawidzim:
With code is all fine. But Riviera-PRO 2020.04 has problem with this fragment

Silabs-ArjanB:
Hi @dawidzim I discussed with @davideschiavone and we would like to keep this repos as simple as possible, specifically not adding back in support for more simulators. We do however appreciate your work and we aim to support as many simulators as possible. The right place for extending simulator support for this core would be in core-v-verif (and I see that you did a pull request there as well). Also, in general, we do not intend to update the RTL for each simulator work-around, so ideally the issue you mentioned gets fixed in the simulator itself.

PR #544, opened 2020-10-07
reimplement tracer to be more resilient to pipeline stalls due to OBI…

strichmo:
… gnt and rvalid stalls

breaks out register file wb and ex updates from pipeline
moves tracer class into an include file for code cleanliness
add wb_bypass bit to signal that gpr check can be relaxed for branch

**This PR must be merged before core-v-verif PR which must point to this**: https://github.com/openhwgroup/core-v-verif/pull/279

### DIFF
--- a/bhv/cv32e40p_instr_trace.svh
+++ b/bhv/cv32e40p_instr_trace.svh
@@ -1,0 +1,864 @@
+typedef struct {
+    logic [ 5:0] addr;
+    logic [31:0] value;
+    bit          filled;
+  } reg_t;
+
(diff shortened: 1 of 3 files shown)


Silabs-ArjanB (changes requested):
Hi @strichmo Can you add a license header to the .svh file, then we are good to go

strichmo:
Thanks @Silabs-ArjanB .  Addressed the concern with latest push.

Silabs-ArjanB (approved):

PR #273, opened 2020-03-23
Initial version of PD files for Cadence flow

ludmila-bta:
Signed-off-by: Ludmila Rubin <lrubin@btadesignservices.com>https://github.com/openhwgroup/cv32e40p/compare/master...ludmila-bta:ludmilar_branch?expand=1#diff-73e8ed6d6a0dacff4a5610c4720f1d73
https://github.com/openhwgroup/cv32e40p/compare/master...ludmila-bta:ludmilar_branch?expand=1#diff-0a2f379bfe791dc3bb9f47b3a041641e
https://github.com/openhwgroup/cv32e40p/compare/master...ludmila-bta:ludmilar_branch?expand=1#diff-9056380be1c7abd53dce002ff63d3633

PR #545, opened 2020-10-08
Added CV32E40P example constraints

Silabs-ArjanB:
Hi @davideschiavone Can you please review these example constraints?

Could you also close/reject the following as I think we should not be adding tool/technology specific flows in this repos (that could be done in core-v-mcu if we have the need for it)

- https://github.com/openhwgroup/cv32e40p/pull/305
- https://github.com/openhwgroup/cv32e40p/pull/273

### DIFF
--- a/constraints/cv32e40p_core.sdc
+++ b/constraints/cv32e40p_core.sdc
@@ -1,0 +1,167 @@
+# Copyright 2020 Silicon Labs, Inc.
+#
+# This file, and derivatives thereof are licensed under the
+# Solderpad License, Version 2.0 (the "License").
+#
+# Use of this file means you agree to the terms and conditions
+# of the license and are in full compliance with the License.
+#
+# You may obtain a copy of the License at:
+#
+#     https://solderpad.org/licenses/SHL-2.0/
+#
+# Unless required by applicable law or agreed to in writing, software
+# and hardware implementations thereof distributed under the License
+# is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS
+# OF ANY KIND, EITHER EXPRESSED OR IMPLIED.
+#
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#//////////////////////////////////////////////////////////////////////////////
+# Engineer:       Arjan Bink - arjan.bink@silabs.com                         //
+#                                                                            //
+# Project Name:   CV32E40P                                                   //
+#                                                                            //
+# Description:    Example synthesis constraints.                             //
+#                                                                            //
+#                 The clock period and input/output delays are technology    //
+#                 and project dependent and are expected to be adjusted as   //
+#                 needed.                                                    //
+#                                                                            //
+#                 OBI related bus inputs arrive late on purpose and OBI      //
+#                 related outputs are available earlier (as they shall not   //
+#                 combinatorially depend o nthe OBI inputs)                  //
+#                                                                            //
+#//////////////////////////////////////////////////////////////////////////////
+
+# 200MHz
+set clock_period 5.0
+
+# Input delays for interrupts
+set in_delay_irq          [expr $clock_period * 0.50] 
+# Output delays for interrupt related signals
+set out_delay_irq         [expr $clock_period * 0.25] 
+
+# Input delays for early signals
+set in_delay_early [expr $clock_period * 0.10] 
+
+# OBI inputs delays
+set in_delay_instr_gnt    [expr $clock_period * 0.83]
+set in_delay_instr_rvalid [expr $clock_period * 0.83]
+set in_delay_instr_rdata  [expr $clock_period * 0.83]
+
+set in_delay_data_gnt     [expr $clock_period * 0.83]
+set in_delay_data_rvalid  [expr $clock_period * 0.83]
+set in_delay_data_rdata   [expr $clock_period * 0.83]
+
+# OBI outputs delays
+set out_delay_instr_req  [expr $clock_period * 0.57]
+set out_delay_instr_addr [expr $clock_period * 0.57]
+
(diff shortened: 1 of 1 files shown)


constraints/cv32e40p_core.sdc
```diff
@@ -0,0 +1,167 @@
...
+
+# OBI outputs delays
+set out_delay_instr_req  [expr $clock_period * 0.57]
+set out_delay_instr_addr [expr $clock_period * 0.57]
+
+set out_delay_data_req   [expr $clock_period * 0.57]
+set out_delay_data_we    [expr $clock_period * 0.57]
+set out_delay_data_be    [expr $clock_period * 0.57]
+set out_delay_data_addr  [expr $clock_period * 0.57]
+set out_delay_data_wdata [expr $clock_period * 0.57]
```
davideschiavone, line 66: Why cannot we assume that output and input delay from/to memories are the same?
These numbers (0.83 and 0.57) look very specific to a given technology

constraints/cv32e40p_core.sdc
```diff
@@ -0,0 +1,167 @@
...
+
+# I/O delays for non RISC-V Bus Interface ports
+set in_delay_other       [expr $clock_period * 0.10]
+set out_delay_other      [expr $clock_period * 0.57]
```
davideschiavone, line 70: same as before

davideschiavone (changes requested):

constraints/cv32e40p_core.sdc
Silabs-ArjanB: Hi @davideschiavone Would it be okayif I just round the numbers: 0.57 -> 0.6, 0.83 -> 0.8. The numbers will always be tech specific, which is why there is a comment at the top about that.

constraints/cv32e40p_core.sdc
Silabs-ArjanB: Hi @davideschiavone I want to convey that memory related output come out roughly midway through the cycle and that memory related inputs are expected to arrive late. The numbers will always be tech specific, which is why there is a comment at the top about that. Would it be okay if I just round the numbers: 0.57 -> 0.6, 0.83 -> 0.8. I do not really care about the actual numbers as long as the signals arriving from memory are assumed to arrive later than that we will guarantee the availability of the signals towards memory (this will assure there are no dis-allowed combinatorial paths inside the core for the OBI interfaces. If you have other nrs in mind, please let me know.

constraints/cv32e40p_core.sdc
Silabs-ArjanB: Okay to just round 0.57 to 0.6?

MikeOpenHWGroup:
Thanks for this PR @Silabs-ArjanB.  Having synthesis constraints adds credibility to this project.   Couldl you add the "SPDX" license tag to the header?
```
# SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
```
I hate to nag about this, but it _is_ important.   Thanks.

Silabs-ArjanB:
> 
> 
> Thanks for this PR @Silabs-ArjanB. Having synthesis constraints adds credibility to this project. Couldl you add the "SPDX" license tag to the header?
> 
> ```
> # SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
> ```
> 
> I hate to nag about this, but it _is_ important. Thanks.

Hi @MikeOpenHWGroup The above is just our standard license file. I just looked at the latest commit in core-v-verif and it looks the same to me, see https://github.com/MikeOpenHWGroup/core-v-verif/blob/1c36bde284a0e9971dc02cf2fa7c1c3940cf2d08/cv32/env/uvme_cv32/uvme_cv32_tdefs.sv

Silabs-ArjanB:
> 
> 
> > I just looked at the latest commit in core-v-verif and it looks the same to me
> 
> You are right about that. I have an action item from the Eclipse Foundation to add the SPDX line to all files in core-v-verif. Sounds like fun doesn't it? Just trying to save you the same hassle.

I propose that somebody writes a script and fixes all files in this repos as opposed to do ad-hoc single fixes.

MikeOpenHWGroup:
> I propose that somebody writes a script and fixes all files in this repos as opposed to do ad-hoc single fixes.

I agree that would be the best (well, at least the fastest) approach.  However, for core-v-verif at least, I need to ensure that the original contributor of the code signs off on any changes to the copyright header.  Painful.

constraints/cv32e40p_core.sdc
davideschiavone: 0.6 and 0.8 with the comments look ok. 
Is there a specific reason why we are saying that we leave inside the core 40% to go out (1-0.6) and we have only 20% (1-0.8) from the response? why not , for example, 30% each?

constraints/cv32e40p_core.sdc
davideschiavone: yes

constraints/cv32e40p_core.sdc
Silabs-ArjanB: The numbers are based on what is possible plus what is expected from the environment. In particular the OBI inputs are expected to arrive very late in the cycle if you are building a system without wait states.

Silabs-ArjanB:
> 
> 
> > I just looked at the latest commit in core-v-verif and it looks the same to me
> 
> You are right about that. I have an action item from the Eclipse Foundation to add the SPDX line to all files in core-v-verif. Sounds like fun doesn't it? Just trying to save you the same hassle.

Yes, sounds like fun indeed. For now I stick to the header that my company approved though.

constraints/cv32e40p_core.sdc
davideschiavone: Yes I understand the number per se, I don't understand why it's different between input and output

PR #559, opened 2020-10-28
fix #363

### LINKED ISSUE #363: Wrong (none) bypass usage by pv.cplxconj
The pv.cplxconj instruction does not correctly use the register file bypasses. In the following example the pv.cplxconj.b should get its x12 operand from the preceding lui instruction. However, the register file bypass is not triggered and the wrong ALU operand is used for the pv.cplxconj.b instruction.

                Time          Cycles PC       Instr    Mnemonic
     86ns               4 00000180 00000013 nop                                 
     96ns               5 00000182 00000013 nop                                 
    106ns               6 00000184 fffe3637 lui              x12, 0xfffe3000     x12=fffe3000
    116ns               7 00000186 5cc19657 pv.cplxconj.b    x12, x3, x12        x12=00000000  x3:00000000 x12:00000000
    126ns               8 0000018a 00000013 nop                                 
    136ns               9 0000018c 00000013 nop                                 
    146ns              10 0000018e 00000013 nop                                 

Wanted behavior is achieved for the following code (where I just inserted some nops so that the RTL would not have to rely on the register file bypass):

.section .text.start
.global _start
.type _start, @function

_start:

.word 0x00010001
lui x12,0xfffe3
nop
nop
nop
.word 0x5cc19657
nop
nop
nop

image

Erroneous behavior is achieved for the following code:

.section .text.start
.global _start
.type _start, @function

_start:

.word 0x00010001
lui x12,0xfffe3
.word 0x5cc19657
nop
nop
nop

image

I believe (but didn't check) that the issue is caused by the following RTL in the 'decoder':

          6'b01011_1: begin // pv.cplxconj.h
            alu_operator_o       = ALU_ABS;
            is_clpx_o            = 1'b1;
            scalar_replication_o = 1'b0;
            **regb_used_o          = 1'b0;**
          end

My guess (without understanding what this instruction is supposed to do) is that regb_used_o needs to be 1.

Also note the related ticket https://github.com/openhwgroup/core-v-docs/issues/95 . The RTL code only mentions pv.cplxconj.h, whereas I think the above RTL handles both .h and .b. The documentation only mentions pv.cplxconj and does not mention the existence of pv.cplxconj.b, pv.cplxconj.h.

davideschiavone:


davideschiavone:
Assigned both @Silabs-ArjanB as he opended the issue and @jm4rtin (leader of the PULP extentions) to have a look at the fix

Silabs-ArjanB (commented):
I would expect an actual RTL fix, not just a tracer fix.

davideschiavone:
What shall the RTL do? it has no bugs..

Silabs-ArjanB:
> 
> 
> What shall the RTL do? it has no bugs..

The issue that you refer to here (i.e. https://github.com/openhwgroup/cv32e40p/issues/363) reports an RTL bug (i.e. wrong execution as shown in the included assembly code)

davideschiavone:
I don't see the wrong execution. I see that you report that the operand B does not contain the right value. I am saying that the operand B does not exist for this operation (as for absolute value for example), thus it was a tracer bug, as it should not show any operand B, as state in the documentation as well :)

The .h .b problem instead has been already discussed and fixed

Silabs-ArjanB:
Hi @davideschiavone The two provided waves show a different result in mem[12] based on whether some nops are present or not; that does not sound okay to me.

davideschiavone:
Hi @Silabs-ArjanB , you are right, I was focused on the mem[12] as operand B, instead of result register and missunderstood the problem. Like this we catched 2. The tracer and the actual RTL one.

The RTL is actually a documentation bug, as this instruction reuses the absolute value ALU operation, thus requires one operand to be 0 (i.e. operand B)
I added the illegal condition in case it's not, so your test will still fail as it has both instr[12] and instr[24:20] different than 0, but I tried with and without the NOP now and I get:

```
   1865ns             182 000001d2 fffe3637 c.lui            x12, 0xfffe3000     x12=fffe3000
   1875ns             183 000001d4 5c018657 pv.cplxconj.h    x12, x3             x12=00002dd8  x3:00002dd8
```

with NOPs
```

   1965ns             192 000001d2 fffe3637 c.lui            x12, 0xfffe3000     x12=fffe3000
   1975ns             193 000001d4 00000013 c.nop                               
   1985ns             194 000001d6 00000013 c.nop                               
   1995ns             195 000001d8 00000013 c.nop                               
   2005ns             196 000001da 5c018657 pv.cplxconj.h    x12, x3             x12=00002dd8  x3:00002dd8

```
as now the src2 field is x0.

Sorry for the confusion

Silabs-ArjanB (approved):

Silabs-ArjanB:
> 
> 
> Hi @Silabs-ArjanB , you are right, I was focused on the mem[12] as operand B, instead of result register and missunderstood the problem. Like this we catched 2. The tracer and the actual RTL one.
> 
> The RTL is actually a documentation bug, as this instruction reuses the absolute value ALU operation, thus requires one operand to be 0 (i.e. operand B)
> I added the illegal condition in case it's not, so your test will still fail as it has both instr[12] and instr[24:20] different than 0, but I tried with and without the NOP now and I get:
> 
> ```
>    1865ns             182 000001d2 fffe3637 c.lui            x12, 0xfffe3000     x12=fffe3000
>    1875ns             183 000001d4 5c018657 pv.cplxconj.h    x12, x3             x12=00002dd8  x3:00002dd8
> ```
> 
> with NOPs
> 
> ```
> 
>    1965ns             192 000001d2 fffe3637 c.lui            x12, 0xfffe3000     x12=fffe3000
>    1975ns             193 000001d4 00000013 c.nop                               
>    1985ns             194 000001d6 00000013 c.nop                               
>    1995ns             195 000001d8 00000013 c.nop                               
>    2005ns             196 000001da 5c018657 pv.cplxconj.h    x12, x3             x12=00002dd8  x3:00002dd8
> ```
> 
> as now the src2 field is x0.
> 
> Sorry for the confusion

I understand now, thanks

PR #560, opened 2020-10-28
Removed fpnew dependency

Silabs-ArjanB:
Removed github dependency on fpnew

fpnew is still cloned if the FPU test case is run, but fpnew is not required to compile the CV32E40P RTL itself. 

The content of rtl/include/cv32e40p_pkg.sv needs to be kept in sync with the content of fpnew/src/fpnew_pkg.sv in case FPU=1 is used.


example_tb/core/Makefile
```diff
@@ -99,10 +100,14 @@ vsim-all: .opt-rtl
 	$(RTLSRC_VOPT_TB_TOP)
 	touch .opt-rtl
 
+# fpnew dependencies
+fpnew_deps:
+	rm -rf ../../rtl/fpnew; git clone https://github.com/pulp-platform/fpnew ../../rtl/fpnew --recurse
```
davideschiavone, line 105: If you run multiple times this commands, it clones everytime the repo. Same when you execute multiple times the custom-fp program. Can the rm -rf part being part of the clean target rather than fpnew_deps?

davideschiavone (changes requested):

example_tb/core/Makefile
Silabs-ArjanB: Hi @davideschiavone I changed the Makfile based on your comment; can you have another look please?

davideschiavone (approved):

davideschiavone:
LGTM, you can merge

PR #570, opened 2020-11-06
fix format string in action block of an assertion

strichmo:


Silabs-ArjanB (approved):

Silabs-ArjanB:
@strichmo Thanks; merged

PR #577, opened 2020-11-11
Added explicit names to generated blocks

Silabs-ArjanB:
Hi @strichmo Can you please have a look and let me know if you agree with the naming style. I tried to give all generated blocks a gen_ prefixed name.

The code is sequentially equivalent to the original code (had to rewrite some code in register file and CSR file to get proper block naming).

Reason for the gen_ prefixing is to have constant names that can be used in waivers, code coverage, etc.

### DIFF
--- a/rtl/cv32e40p_alu.sv
+++ b/rtl/cv32e40p_alu.sv
@@ -61,7 +61,7 @@ module cv32e40p_alu import cv32e40p_pkg::*;
   generate
     genvar k;
     for(k = 0; k < 32; k++)
-    begin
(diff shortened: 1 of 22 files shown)


strichmo (approved):
Thanks for the changes.  These look fine to me.

strichmo:
I have pulled this locally and am 1/4 through a full regression.  I would've seen any failures by now so I am good with merging this PR.
I will need to update vRefine files but I can do that later (post-merge) in core-v-verif.

PR #578, opened 2020-11-11
remove RTL code from cv32e40p_controller that is unreachable

silabs-PaulZ:
The following code in the FLUSH_WB state was proven to be unreachable with a formal tool and removed. (no issue reported)

image


Silabs-ArjanB (approved):
Looks good

PR #580, opened 2020-11-13
remove code that is unreachable in cv32e40p_controller FSM. 

silabs-PaulZ:
In particular, DBG_FLUSH -> DBG_FLUSH cannot occur.

image

### DIFF
--- a/rtl/cv32e40p_controller.sv
+++ b/rtl/cv32e40p_controller.sv
@@ -1225,6 +1225,7 @@ module cv32e40p_controller import cv32e40p_pkg::*;
               ctrl_fsm_ns = DBG_TAKEN_ID;
             end else
             begin
+              // else must be debug_single_step_i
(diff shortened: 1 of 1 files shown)


rtl/cv32e40p_controller.sv
```diff
@@ -1223,7 +1223,8 @@ module cv32e40p_controller import cv32e40p_pkg::*;
              debug_req_entry_q                     )
             begin
               ctrl_fsm_ns = DBG_TAKEN_ID;
-            end else if (debug_single_step_i) begin
+            end else
+            begin
```
Silabs-ArjanB, line 1227: Hi @silabs-PaulZ Can you add a comment here that the else branch takes care of 'single stepping'?

Silabs-ArjanB (changes requested):
See inline request to add a comment

Silabs-ArjanB (approved):

PR #582, opened 2020-11-16
:art: changing to enumerator for alu opcodes

davideschiavone:
Hi @Silabs-ArjanB , with this PR the ALU is now showing in the waveform the name of the operation rather than it's value. Making debug easier.

It should be LEC as it's just a style change. If you like it, I make it also for the multiplier and CSR operation


Silabs-ArjanB (approved):

davideschiavone:
Hi @Silabs-ArjanB  - I updated directly this PR with the MUL and CSR opcodes - Let me know whether I can merge

Silabs-ArjanB (approved):

PR #596, opened 2020-12-02
fix for DC Shell

davideschiavone:
IDLE is defined both in the prefetcher FSM and the MULT FSM.

DC Shell does not like that as:

```
dc_shell> analyze -format sverilog  -work work ${DESIGN_RTL_DIR}/cv32e40p_mult.sv
Running PRESTO HDLC
Compiling source file ../../riscv/rtl//cv32e40p_mult.sv
Error:  ../../riscv/rtl//cv32e40p_mult.sv:96: redeclaration of symbol IDLE as constant. (VER-513)
Error:  ../../riscv/rtl//cv32e40p_mult.sv:96: symbol IDLE must be a constant or parameter. (VER-260)
* Presto compilation terminated with 2 errors. *
```

This PR solves it

### DIFF
--- a/rtl/cv32e40p_mult.sv
+++ b/rtl/cv32e40p_mult.sv
@@ -136,7 +136,7 @@ module cv32e40p_mult import cv32e40p_pkg::*;
     multicycle_o     = 1'b0;
 
     case (mulh_CS)
-      BEGIN: begin
+      IDLE_MULT: begin
         mulh_active = 1'b0;
         mulh_ready  = 1'b1;
         mulh_save   = 1'b0;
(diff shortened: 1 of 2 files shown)


rtl/cv32e40p_mult.sv
```diff
@@ -138,7 +136,7 @@ module cv32e40p_mult import cv32e40p_pkg::*;
     multicycle_o     = 1'b0;
 
     case (mulh_CS)
-      IDLE: begin
+      BEGIN: begin
```
Silabs-ArjanB, line 139: It is not okay to use a (capitalized) Verilog/VHDL keyword as a constant; please come up with a new name. You could just keep using IDLE; you just have to prefix its usage with the proper type def.

Silabs-ArjanB (changes requested):
Please do not use a Verilog/VHDL keyword.

rtl/cv32e40p_mult.sv
davideschiavone: you are right @Silabs-ArjanB !!!

Silabs-ArjanB (approved):

PR #597, opened 2020-12-02
Add missing license headers

bluewww:
Added license headers for files that are under ETH copyright. Technically already covered by the `LICENSE` file in the root folder, but this makes it more clear

### DIFF
--- a/example_tb/core/firmware/stats.c
+++ b/example_tb/core/firmware/stats.c
@@ -1,9 +1,19 @@
-// This is free and unencumbered software released into the public domain.
-//
-// Anyone is free to copy, modify, publish, use, compile, sell, or
-// distribute this software, either in source code form or as a compiled
-// binary, for any purpose, commercial or non-commercial, and by any
-// means.
+/*
+ * Copyright 2020 ETH Zurich
+ *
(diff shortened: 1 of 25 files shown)


example_tb/core/firmware/stats.c
```diff
@@ -1,9 +1,19 @@
-// This is free and unencumbered software released into the public domain.
```
davideschiavone, line 1: are you sure we can remove this piece of text? @bluewww

example_tb/core/interrupt/isr.h
```diff
@@ -1,9 +1,19 @@
-// This is free and unencumbered software released into the public domain.
```
davideschiavone, line 1: as before @bluewww

davideschiavone:
amo_shim.sv should also be under solderpad and software.tcl under apache @bluewww ! Thanks a lot

example_tb/core/interrupt/isr.h
bluewww: I think this is fine. It says in the text that you can modify this code however you want

bluewww:
Updated

Issue #609, opened 2021-01-25
Synopsys error


timsaxe:
# Template for Bug Issue

Please use this template if the issue that you are reporting is a (potential) bug in CV32E40P RTL, Documentation, Verification environment or Tool and Build system.

<br>**Note**: Please file bugs related to the CV32E40P User Manual in https://github.com/openhwgroup/core-v-docs instead.
<br>**Note**: Please file bugs related to the core-v-verif verification environment in https://github.com/openhwgroup/core-v-verif instead.
<br>**Note**: Please file bugs related to the PULP tool chain in https://github.com/pulp-platform/pulp-riscv-gnu-toolchain instead.

## Bug Title
When synthesizing cv32e40p Synopsys reports the following:

Compiling source file ../../ips/cv32e40p/./rtl/cv32e40p_pmp.sv
Error:  ../../ips/cv32e40p/./rtl/cv32e40p_pmp.sv:671: redeclaration of symbol IDLE as constant. (VER-513)
Error:  ../../ips/cv32e40p/./rtl/cv32e40p_pmp.sv:671: symbol IDLE must be a constant or parameter. (VER-260)
*** Presto compilation terminated with 2 errors. ***

### Component
Indicate the component in which you found a problem (and we will add the appropriate label):

**Component:RTL**: For issues in the RTL (e.g. for files in the rtl directory)

### Steps to Reproduce
Please provide:
Using tag _v1.0.0

Search in the relevant files reports:
Search "IDLE" (6 hits in 2 files of 3 searched)
  C:\Users\saxe\Downloads\cv32e40p_pkg.sv (2 hits)
	Line 206: typedef enum logic {IDLE, BRANCH_WAIT} prefetch_state_e;
	Line 208: typedef enum logic [2:0] {IDLE_MULT, STEP0, STEP1, STEP2, FINISH} mult_state_e;
  C:\Users\saxe\Downloads\cv32e40p_pmp.sv (4 hits)
	Line 671:    enum logic {IDLE, GIVE_ERROR} data_err_state_q, data_err_state_n;
	Line 679:          IDLE:
	Line 689:                data_err_state_n = IDLE;
	Line 697:           data_err_state_q <= IDLE;

If I change 'enum logic {IDLE, GIVE_ERROR} data_err_state_q, data_err_state_n;' to
'enum logic {IDLE_ERROR, GIVE_ERROR} data_err_state_q, data_err_state_n;'
[...]

PR #610, opened 2021-01-25
remove PMP from src file

davideschiavone:
It removes the PMP for #609 and updates the src file to use the FF reg file.
It includes the PR to be accepted #604

This PR should be merged after #604 and after having a LEC CI in place

### DIFF
--- a/src_files.yml
+++ b/src_files.yml
@@ -22,13 +22,10 @@ cv32e40p_regfile_rtl:
     ./rtl/include,
   ]
   files: [
-    ./rtl/cv32e40p_register_file_latch.sv,
+    ./rtl/cv32e40p_register_file_ff.sv,
   ]
 
 cv32e40p:
-  vlog_opts: [
-    -L fpnew_lib,
-  ]
   incdirs: [
     ./rtl/include,
     ../../rtl/includes,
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup (changes requested):
The core-v-mcu should not be using its own manifest for the cv32e40p.  This should be read in directly from a manifest in the cv32e40p repo.

Silabs-ArjanB (approved):

davideschiavone:
> 
> 
> The core-v-mcu should not be using its own manifest for the cv32e40p. This should be read in directly from a manifest in the cv32e40p repo.

Can you please open an issue on the core-v-mcu repository then? we decided to maintain the src_file and Bender too for the PULP team. If we want to use the manifest in the current pulp-based openHW MCU we should open an issue against that in that repo. In the core side there is nothing wrong. Thanks for pointing it out!

MikeOpenHWGroup:
> Can you please open an issue on the core-v-mcu repository then?

Will do @davideschiavone.  Thanks.

Silabs-ArjanB (approved):

MikeOpenHWGroup (approved):

PR #611, opened 2021-01-26
Remove cv32e40p pmp from synth

timsaxe:
This file causes Synopsys to fail (doesn't like redefinitiion of IDLE) and the documentation says that it is  not required for synthesis.

### DIFF
--- a/src_files.yml
+++ b/src_files.yml
@@ -35,6 +35,7 @@ cv32e40p:
   ]
   files: [
     ./rtl/include/cv32e40p_apu_core_pkg.sv,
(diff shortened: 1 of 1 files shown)


davideschiavone:
Hi @timsaxe , I have already made a PR at #610  so I close this issue ok?
I should also add the FPU package. I also replaced the Latch based RF with the flip-flop one. As that one is the verified one.
As @MikeOpenHWGroup pointed out, the best would be using the manifest file in the PULP flow.

timsaxe:
Yes, please close.

PR #623, opened 2021-02-18
delay tracer now longer than randomstall

davideschiavone:
When simulating the example_tb with modelsim with random stalls on Data and Instruction memory as:

`make custom-vsim-run CUSTOM_GCC_FLAGS="-DRANDOM_MEM_STALL"
`
I get the errors:

UVM_ERROR ../../bhv/cv32e40p_tracer.sv(443) @ 3085ns: reporter [TRACER] WB .....

Intuitively, as the random stalls add delay, I though it was due a race condition.
Therefore I made the time of the random stall generator shorter and the one from the tracer longer.

This solved my issue.

Let me know if that makes sense

Signed-off-by: Pasquale Davide Schiavone


Silabs-ArjanB (commented):
I really can't comment on this as I don't know the scenarios for which these # delays were introduced. It looks quite dirty of course. At first glance I would say that the tracer itself should not be changed as that might then cause similar issues in core-v-verif. Maybe you can check how the stall functionality is implemented there and do the same in example_tb.

PR #625, opened 2021-02-23
Fix tracer during random gnt and valid stalls

davideschiavone:
When simulating the example_tb with modelsim with random stalls on Data and Instruction memory as:

make custom-vsim-run CUSTOM_GCC_FLAGS="-DRANDOM_MEM_STALL"
I get the errors:

UVM_ERROR ../../bhv/cv32e40p_tracer.sv(443) @ 3085ns: reporter [TRACER] WB .....

I imported from the core-v-verif the modules to create such stalls as suggested in #623 

Plus, I updated the Makefile to support xpulp instructions

Signed-off-by: Pasquale Davide Schiavone <davide@openhwgroup.org>


example_tb/core/riscv_gnt_stall.sv
```diff
@@ -0,0 +1,144 @@
...
+  else begin
+`ifdef VERILATOR
+    //#1;
+`else
+    #(100ps);
+`endif
+
+    // When request is removed, remove grant
+    if (!req_core_i) begin
+      grant_core_o <= 1'b0;
```
Silabs-ArjanB, line 118: This limits coverage. I will file an issue in core-v-verif instead, no need to make changes here (yet)

Silabs-ArjanB (approved):

PR #635, opened 2021-04-12
fixed #616

### LINKED ISSUE #616: Document immediate sign-extend for PULP post-increment loads/stores
For the following instructions the documentation does not indicate that the immediate is sign-extended:

- p.lb rD, Imm(rs1!)       
- p.lbu rD, Imm(rs1!)      
- p.lh rD, Imm(rs1!)       
- p.lhu rD, Imm(rs1!)      
- p.lw rD, Imm(rs1!)       
- p.sb rs2, Imm(rs1!)      
- p.sh rs2, Imm(rs1!)      
- p.sw rs2, Imm(rs1!)      

My proposal would be to replace 'rs1 += Imm[11:0]' by 'rs1 += Sext(Imm[11:0])' in their descriptions.

davideschiavone:
Fix issue #616 

As this is a bug in the doc, it may lead to erros in the compiler porting as well. Thus, pinging @jeremybennett as well

### DIFF
--- a/docs/source/instruction_set_extensions.rst
+++ b/docs/source/instruction_set_extensions.rst
@@ -1,18 +1,18 @@
 ..
    Copyright (c) 2020 OpenHW Group
-   
+
    Licensed under the Solderpad Hardware Licence, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
    You may obtain a copy of the License at
-  
+
    https://solderpad.org/licenses/
-  
+
    Unless required by applicable law or agreed to in writing, software
    distributed under the License is distributed on an "AS IS" BASIS,
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
-  
+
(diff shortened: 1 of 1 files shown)


Silabs-ArjanB (approved):

davideschiavone:
hi @jeremybennett , did you have time to have a look at it?

jeremybennett (approved):
This much more needs advice from @jessicamills and the GNU tools team. In particular the work on the PULP load/store extensions has been done by @NBruschi and @EnricoTabanelli assisted by @gtagliavini

My immediate thoughts are that the work to date has been primarily on the assembler and builtins. For the assembler the semantics of the instruction don't matter, but we need to ensure that signed literals are permitted in the syntax for the assembler. For the builtins it will affect whether signed or unsigned types are used for parameters.

There may be impacts on the linker if there are any relocations impacted - for example if the index were symbolic, then potentially we need a signed rather than unsigned relocation.

I don't believe we have any committed work on GCC, where this will be more significant. However I think @NBruschi and @EnricoTabanelli have been working on this, and should comment on the impact.

I'm happy to approve, since this comment is about ensuring that those impacted in the tool chain have an opportunity to consider the implications for their existing and future work.

PR #638, opened 2021-04-27
changing RTL to be compliant with verible

davideschiavone:
Signed-off-by: davideschiavone <davide@openhwgroup.org>


davideschiavone:
@Silabs-ArjanB , I ran the LEC script before submitting this PR

Silabs-ArjanB (approved):
Okay with this as it is LEC equivalent

PR #658, opened 2021-08-20
fix #657

### LINKED ISSUE #657: cv.add.{div2,div4,div8} encoding in DOC does not match RTL
In the documentation, 

cv.add.{div2,div4,div8} func5 (bits 31-27) are: **01011**

However, the RTL is: **01110**

This is an error of the Documentation that may effect the compiler too

davideschiavone:
@nbruschi can you have a look at it?


NBruschi:
@davideschiavone all looks good.

PR #675, opened 2021-11-10
Update debug.rst

### LINKED ISSUE #669: Please update the User Manual to clarify "interrupt occurs during step" behavior
As discussed in [core-v-verif issue 904](https://github.com/openhwgroup/core-v-verif/issues/904), the specified behaviour in Debug Spec 0.13.2 for a RISC-V core when an interrupt occurs while the core is stepping is somewhat ambiguous.  After the 0.13.2 spec was released, this ambiguity was discussed in:

 https://github.com/riscv/riscv-debug-spec/issues/510
 https://github.com/riscv/riscv-debug-spec/issues/511

Debug Spec 1.0.0 clarifies the expected behaviour of a core (Section 4.1.1).  The CV32E40P is _not_ compliant with this section of the 1.0.0 spec, and it was agreed that the CV32E40P User Manual should be updated to reflect this.  I leave it to others to determine whether this should be classified as an errata against 1.0.0, or a clarification against 0.13.2.

Historical context is provided in the attachment:
[cv32e40_debug_issue(14-oct-2021).pdf](https://github.com/openhwgroup/cv32e40p/files/7346559/cv32e40_debug_issue.14-oct-2021.pdf)

davideschiavone:
This should clarify interrupts during single-step to fix #669

### DIFF
--- a/docs/source/debug.rst
+++ b/docs/source/debug.rst
@@ -185,8 +185,8 @@ Execuitng the EBREAK instruction when the core is in Debug Mode shall result in 
 Interrupts during Single-Step Behavior
 --------------------------------------
 
-The CV32E40P CPU is compliant with the 0.13.2 RISC-V Debug spec.
-The RISC-V Debug spec specified what the behavior of interrupts occurring during Single-Step must be only in version 1.0.0, after the implementation of the CV32E40P core. 
+The CV32E40P CPU is not compliant with the intended interpretation of the RISC-V Debug spec 0.13.2 specification when interrupts occur during Single-Steps. 
+However, the intended behavior has been clarified a posteriori only in version 1.0.0.
 See https://github.com/riscv/riscv-debug-spec/issues/510. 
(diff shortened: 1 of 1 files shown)


docs/source/debug.rst
```diff
@@ -178,3 +178,15 @@ Execuitng the EBREAK instruction when the core is in Debug Mode shall result in
...
 - The core remains in Debug Mode and execution jumps back to the beginning of the debug code located at ``dm_halt_addr_i``
 - none of the CSRs are modified
+
+
+.. _interrupts_single-stepping:
+
+Interrupts during Single-Step Behavior
+--------------------------------------
+
+The CV32E40P CPU is compliant with the 0.13.2 RISC-V Debug spec.
```
Silabs-ArjanB, line 188: I think this is a bit too strong. I would say that the 40P is not compliant to the intended interpretation of the 0.13.2 specification, but that this intended behavior was only clarified in version 1.0.0.

davideschiavone: agree and fixed

davideschiavone: let's wait for @strichmo before merging in

PR #680, opened 2022-02-14
vendorized fpu

davideschiavone:
Added local copy of FPU and sub dependencies


davideschiavone:
The manifest file should be updated in a second PR - suggestions needed

Silabs-ArjanB (commented):
Why can't the FPU be put into its own repos? I don't think it belongs in the cv32e40p repos.

pascalgouedo:
Hi
To keep coherency with core-v-mcu and have all rtl in the same place, maybe it is better to put vendor directory in rtl.

Silabs-ArjanB (approved):

davideschiavone:
thanks - waiting for @MikeOpenHWGroup

MikeOpenHWGroup (approved):

PR #683, opened 2022-03-08
Fix tracer for FPU

bluewww:
This PR addresses two issues:
Firstly,
when the FPU latency in `cv32e40p_pkg.sv` is changed to e.g.
```diff
-  parameter int unsigned C_LAT_FP32 = 'd0;
+  parameter int unsigned C_LAT_FP32 = 'd1;
```
then tracer will hang or write x's to the trace log. This is due to the tracer not correctly waiting for the APU handshaking to start and conclude in the execution stage. We fix this by introducing an execution delay stage to model these multi-cycle APU instructions.

and secondly when ZFINX is enabled, it still looks like as if the fpu instructions are referring to non-existing floating point registers. This is just a cosmetic issue.

### DIFF
--- a/bhv/cv32e40p_instr_trace.svh
+++ b/bhv/cv32e40p_instr_trace.svh
@@ -61,7 +61,7 @@ class instr_trace_t;
   reg_t        regs_read[$];
   reg_t        regs_write[$];
   mem_acc_t    mem_access[$];
-  logic        retired;
+  logic        is_apu;
 
   function new();
     str        = "";
--- a/bhv/cv32e40p_instr_trace.svh
+++ b/bhv/cv32e40p_instr_trace.svh
@@ -389,9 +389,14 @@ class instr_trace_t;
       regs_read.push_back('{rs2, rs2_value, 0});
       regs_read.push_back('{rs4, rs3_value, 0});
       regs_write.push_back('{rd, 'x, 0});
-      str = $sformatf(
-          "%-16s f%0d, f%0d, f%0d, f%0d", mnemonic, rd - 32, rs1 - 32, rs2 - 32, rs4 - 32
-      );
+      if (PULP_ZFINX)
+        str = $sformatf(
+            "%-16s x%0d, x%0d, x%0d, x%0d", mnemonic, rd, rs1, rs2, rs4
+        );
+      else
+        str = $sformatf(
+            "%-16s f%0d, f%0d, f%0d, f%0d", mnemonic, rd - 32, rs1 - 32, rs2 - 32, rs4 - 32
+        );
     end
   endfunction  // printF3Instr
 
@@ -400,7 +405,10 @@ class instr_trace_t;
       regs_read.push_back('{rs1, rs1_value, 0});
       regs_read.push_back('{rs2, rs2_value, 0});
       regs_write.push_back('{rd, 'x, 0});
-      str = $sformatf("%-16s f%0d, f%0d, f%0d", mnemonic, rd - 32, rs1 - 32, rs2 - 32);
+      if (PULP_ZFINX)
+        str = $sformatf("%-16s x%0d, x%0d, x%0d", mnemonic, rd, rs1, rs2);
+      else
+        str = $sformatf("%-16s f%0d, f%0d, f%0d", mnemonic, rd - 32, rs1 - 32, rs2 - 32);
     end
   endfunction  // printF2Instr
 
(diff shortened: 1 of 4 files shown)


davideschiavone:
thanks a lot @bluewww - I let @pascalgouedo to review it as the owner of the FPU related changes in the core

abukharmeh:
@sarkuuk We will likely have to pick these tracer changes to the 41p

bhv/cv32e40p_tracer.sv
```diff
@@ -176,20 +189,21 @@ module cv32e40p_tracer
...
-  assign rd  = {rd_is_fp,  instr[11:07]};
-  assign rs1 = {rs1_is_fp, instr[19:15]};
-  assign rs2 = {rs2_is_fp, instr[24:20]};
-  assign rs3 = {rs3_is_fp, instr[29:25]};
-  assign rs4 = {rs3_is_fp, instr[31:27]};
+  assign rd  = {PULP_ZFINX ? 0 : rd_is_fp,  instr[11:07]};
+  assign rs1 = {PULP_ZFINX ? 0 : rs1_is_fp, instr[19:15]};
+  assign rs2 = {PULP_ZFINX ? 0 : rs2_is_fp, instr[24:20]};
+  assign rs3 = {PULP_ZFINX ? 0 : rs3_is_fp, instr[29:25]};
+  assign rs4 = {PULP_ZFINX ? 0 : rs3_is_fp, instr[31:27]};
```
abukharmeh, line 206: I wonder why is this necessary, the values for r*_is_fp should be always correct regardless?

davideschiavone: indeed it's probably better to add this extra condition as I quickly saw these signals coming from the decoder only depending on the FPU parameter and not on the PULP_ZFINX

davideschiavone: @bluewww or @abukharmeh, can you add to the PR this check?

abukharmeh: I personally think this should be fixed in the decoder not in the tracer. Apparently the register file ignored the destination whenever ZFINX was set. IMHO, think this will make the code less readable for no apparent saving !

davideschiavone: Agree - but this should not be part of this PR - can you open an issue about it pls?

abukharmeh: Sure, I will do, and will also fix it in next iteration of 41p.

pascalgouedo:
First commit (7fdd45f) for Tracer not necessary any more with #684/#685.
But really interested to use/test the second one.

bhv/cv32e40p_instr_trace.svh
```diff
@@ -389,9 +389,14 @@ class instr_trace_t;
...
-      );
+      if (PULP_ZFINX)
+        str = $sformatf(
+            "%-16s x%0d, x%0d, x%0d, x%0d", mnemonic, rd, rs1, rs2, rs4
+        );
+      else
+        str = $sformatf(
+            "%-16s f%0d, f%0d, f%0d, f%0d", mnemonic, rd - 32, rs1 - 32, rs2 - 32, rs4 - 32
+        );
     end
```
abukharmeh, line 400: Now that we have reg_fp* signals set correctly, we can have some function to retrieve the register name, and get rid of all these if PULP_ZFINX cases :)
```
function string reg_name (logic [5:0] reg_id);
    return reg_id[5] ? $sformatf("f%0d", reg_id[4:0]) : $sformatf("x%0d", reg_id[4:0]);
endfunction
```

davideschiavone: good idea - leaving to @pascalgouedo to evaluate it and include it in his PR

abukharmeh: Little did I know that such function already existed and is configurable to print either ABI names or normal names !
https://github.com/openhwgroup/cv32e40p/blob/1720e8493ccf8617767fc5cb0494d6f6f7bec5b6/bhv/cv32e40p_instr_trace.svh#L282  I think we should use that.

abukharmeh: Yes, sounds good.

abukharmeh: > So maybe SymbolicRegs should be forced to 0 when PULP_ZFINX = 1

I think they can still be meaningful if somebody decided that they want the trace printed with ABI names, I dont know why the original tracer was printing the instruction trace always using normal names, and register writes/reads either either with normal or ABI names depending on SymbolicRegs, I think it make sense that its consistent.

pascalgouedo (approved):
#687 improves and replaces this PR so it could be dropped.

bluewww:
agreed, closing this

PR #685, opened 2022-03-14
Fix #684

### LINKED ISSUE #684: reg_fp* values are wrong when PULP_Zfinx is enabled
https://github.com/openhwgroup/cv32e40p/blob/b7c91867b2f611013dbe98150e745b1a2ff6b7c6/rtl/cv32e40p_decoder.sv#L1412-L1414

The values of reg_fp_* are assigned incorrectly when Zfinx is enabled, and then overridden in Register file, and now in #683  in the tracer. 

I think PULP_ZFINX parameter value should be passed to the decoder module, and these values should be assigned to `!PULP_ZFINX`

pascalgouedo:

### DIFF
--- a/rtl/cv32e40p_decoder.sv
+++ b/rtl/cv32e40p_decoder.sv
@@ -28,9 +28,10 @@
 module cv32e40p_decoder import cv32e40p_pkg::*; import cv32e40p_apu_core_pkg::*; import cv32e40p_fpu_pkg::*;
 #(
   parameter PULP_XPULP        = 1,              // PULP ISA Extension (including PULP specific CSRs and hardware loop, excluding p.elw)
-  parameter PULP_CLUSTER      =  0,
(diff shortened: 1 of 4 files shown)


abukharmeh:
This looks really good, thank you very much.

I wonder if we could do direct assignment from parameter to logic, i.e reg_fp_a_o = 1'(!ZFINX_EN)

I had a quick look at https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md and could not find anything that discourage that ?

davideschiavone:
Hi @pascalgouedo - can you pls send the PR to the dev branch? we now have in place the pipeline to check the logical equivalence on the AWS severs automatically when having a PR on the dev branch
see this example: https://github.com/openhwgroup/cv32e40p/pull/691

davideschiavone (approved):

PR #686, opened 2022-03-23
[LEC] golden reference git url

suppamax:
Signed-off-by: Massimiliano Giacometti <max@openhwgroup.org>


davideschiavone (approved):

davideschiavone:
@Silabs-ArjanB - what happens when something that is neither DOC nor RTL get changed in the PR LABEL?

Silabs-ArjanB:
> @Silabs-ArjanB - what happens when something that is neither DOC nor RTL get changed in the PR LABEL? - shall I used Ignore for release?

If neither doc or rtl is changed, then simply do not apply a label and it won't end up in the changelog.

PR #687, opened 2022-03-23
Fix #683

pascalgouedo:
Hi,

Here are all updates to have correct trace file with respect to FPU instructions (no Xs any more).
It is able to cope with pipelines effects and different fixed latencies FPU instructions (verified in the range 0 to 5 cycles) as well as with variable latencies Integer MUL/DIV and FDIV/FSQRT instructions.

### DIFF
--- a/bhv/cv32e40p_tracer.sv
+++ b/bhv/cv32e40p_tracer.sv
@@ -29,7 +29,10 @@
 module cv32e40p_tracer
   import cv32e40p_pkg::*;
   import uvm_pkg::*;
-(
+#(
+  parameter FPU = 0,
+  parameter PULP_ZFINX = 0
+) (
   // Clock and Reset
   input  logic        clk_i,
   input  logic        rst_n,
@@ -117,7 +120,7 @@ module cv32e40p_tracer
 
 `include "cv32e40p_instr_trace.svh"
 
-  string info_tag = "TRACER";
+  string info_tag;
 
   event         retire;
 
@@ -166,6 +169,8 @@ module cv32e40p_tracer
   initial begin
     wait(rst_n == 1'b1);
     $sformat(fn, "trace_core_%h.log", hart_id_i);
+    $sformat(info_tag, "CORE_TRACER %2d", hart_id_i);
+    $display("[%s] Output filename is: %s", info_tag, fn);
     f = $fopen(fn, "w");
     $fwrite(f, "Time\tCycle\tPC\tInstr\tDecoded instruction\tRegister and memory contents\n");
   end
--- a/bhv/cv32e40p_tracer.sv
+++ b/bhv/cv32e40p_tracer.sv
@@ -99,6 +99,8 @@ module cv32e40p_tracer
   input  logic [ 4:0] imm_clip_type,
 
   input logic         apu_en_i,
+  input logic         apu_singlecycle_i,
+  input logic         apu_multicycle_i,
   input logic         apu_rvalid_i
 
 );
@@ -106,8 +108,9 @@ module cv32e40p_tracer
   import cv32e40p_tracer_pkg::*;
 
   // Make clock a bit to avoid x->0 transitions in tracer logic
-  bit clk_i_d;
-  assign #0.01 clk_i_d = clk_i;
+  bit clk_i_d, eval_comb;
+  assign #0.5 clk_i_d = clk_i;
+  assign eval_comb = ~(clk_i & ~clk_i_d);
 
   event ovp_retire;
   bit  use_iss;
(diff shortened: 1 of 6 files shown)


davideschiavone:
@MikeOpenHWGroup - as this touches the tracer, is it to have a quick running of the step&compare to see if everything is ok? of course I mean the old step&compare, and not the one that includes F instructions as it is not ready yet

bhv/cv32e40p_tracer.sv
```diff
@@ -99,15 +99,18 @@ module cv32e40p_tracer
...
 
 );
 
   import cv32e40p_tracer_pkg::*;
 
   // Make clock a bit to avoid x->0 transitions in tracer logic
-  bit clk_i_d;
-  assign #0.01 clk_i_d = clk_i;
+  bit clk_i_d, eval_comb;
+  assign #0.5 clk_i_d = clk_i;
```
davideschiavone, line 112: why is the delay increased from 0.01 to 0.5?

davideschiavone: I fear it is not a good idea. the test bench may rely on this. In general is never a good thing to use such trick of delaying as we had already, I don't like it, but we have it. I would recommend to not change it to 0.5 if there's not a good reason to do it

MikeOpenHWGroup: !
It is surprising to see this here in the first place, regardless of the delay.  Can it be removed altogether?

pascalgouedo: > I fear it is not a good idea. the test bench may rely on this. In general is never a good thing to use such trick of delaying as we had already, I don't like it, but we have it. I would recommend to not change it to 0.5 if there's not a good reason to do it

The change of time is just for waves. With 10 ps it is hardly visible when debugging. But it could be put back if this is really killer to have 0.5 ns ...

> ! It is surprising to see this here in the first place, regardless of the delay. Can it be removed altogether?

It was in ETH code originally and I used it to generate an eval_comb signal because it seems "always @*" does not work quite well with classes (at least with questa).
I lost a lot of time to understand that the tracer only always comb @* process was not fired on trace_ex/ex_delay/wb changes. I used this eval_comb trick but any other solution resolving this problem is welcome.

Silabs-ArjanB (approved):

bhv/cv32e40p_tracer.sv
```diff
@@ -395,32 +440,66 @@ module cv32e40p_tracer
...
+        move_trace_ex_to_trace_ex_delay = 1;
+        clear_trace_ex              = 1;
+      end else begin
+        move_trace_ex_to_trace_wb = 1;
+        clear_trace_ex            = 1;
+      end
+    end
+
+    if (move_trace_ex_to_trace_wb && move_trace_ex_delay_to_trace_wb) begin
+      $error("ex delay stage and ex stage collide");
```
abukharmeh, line 502: The 40p code base seems to use uvm_error everywhere, but two instances where error is used while checking instantiation parameters to prevent compilation, perhaps this should be also uvm_error ?

bhv/cv32e40p_tracer.sv
```diff
@@ -110,39 +120,47 @@ module cv32e40p_tracer
...
+  logic [31:0] pc_ex_stage;
+  logic [31:0] pc_ex_delay_stage;
+  logic [31:0] pc_wb_stage;
   logic [31:0] pc_wb_delay_stage;
   logic [31:0] pc_retire_head_q;
 
 `include "cv32e40p_instr_trace.svh"
 
-  string info_tag = "TRACER";
+  string info_tag;
```
abukharmeh, line 131: Why did we remove the tag info from uvm messages ?

bhv/cv32e40p_tracer.sv
abukharmeh: This feels like zero delay `(#0)`, i.e trying to move something from active to inactive evaluation region to ensure no race condition occurs. Though, I think this very statment could cause other race conditions and is not particularly safe, but I have no other solution without looking into this further, so I agree that its best to leave as is.

abukharmeh: The eval comb trick is really using NBA to retrigger active region evaluation, I didn't notice it earlier, but this really proves that #0 delay is a bad idea.

bhv/cv32e40p_tracer.sv
bluewww: Oops yes this is probably better handled as uvm error

bhv/cv32e40p_tracer.sv
```diff
@@ -109,7 +109,7 @@ module cv32e40p_tracer
 
   // Make clock a bit to avoid x->0 transitions in tracer logic
   bit clk_i_d, eval_comb;
-  assign #0.5 clk_i_d = clk_i;
+  assign #0.01 clk_i_d = clk_i;
```
bluewww, line 112: What is this even for? Most simulators give me a warning here (for example vcs) that they round this to zero.

bluewww:
Does this replace https://github.com/openhwgroup/cv32e40p/pull/683?

bhv/cv32e40p_tracer.sv
pascalgouedo: Changed to uvm_error

bhv/cv32e40p_tracer.sv
pascalgouedo: Because redefined in initial with additional core nb info.

bhv/cv32e40p_tracer.sv
pascalgouedo: Not rounded in Questa at least.

pascalgouedo:
> Does this replace #683?

Starting from #683, I used @abukharmeh and @davideschiavone feedbacks I agreed with to rework first part of the commit (Add support for ZFINX in tracer) and to centralize updates in cv32e40p_ex_stage.sv.
The second #683 commit (Fix tracer for APU (FPU) instructions) is the starting point of the second part of this #687 as it can be seen in above history. But as it was not working once different FPU latencies were used (up to 5 cycles) it needed additional updates to make it work.

So yes this #687 upgrades #683.

bluewww (approved):
lgtm

Silabs-ArjanB:
> @Silabs-ArjanB shall I tag it to RTL or leave BHV changes without any tag ?

This PR touches RTL files so you would apply the RTL tag.

davideschiavone:
> > @Silabs-ArjanB shall I tag it to RTL or leave BHV changes without any tag ?
> 
> This PR touches RTL files so you would apply the COMPONENT:RTL label.

it's only adding spaces - but I agree with that

PR #717, opened 2022-08-03
Adding support for RVFI

YoannPruvost:
First step toward full RVFI support on the cv32e40p.
For now, only the necessary signals for the ci test are correctly populated


bhv/cv32e40p_tb_wrapper.sv
```diff
@@ -192,6 +194,120 @@ module cv32e40p_tb_wrapper #(
   );
 `endif
 
+
+  cv32e40p_rvfi rvfi_i (
```
davideschiavone, line 198: hi @YoannPruvost , thanks for your contribution.

Can you pls add an ifdef to include or not this module ?

not all the EDA tools will support this module

davideschiavone (changes requested):

bhv/cv32e40p_tb_wrapper.sv
YoannPruvost: Hi @davideschiavone it's done

davideschiavone (approved):

PR #720, opened 2022-09-07
Docs corrections and clarifications

pascalgouedo:
Taking into account some closed or not complete PR.


docs/source/instruction_set_extensions.rst
```diff
@@ -586,10 +586,10 @@ General ALU Operations
 +=================+=========================+========================================================================+
 | **cv.abs**      | **rD, rs1**             | rD = rs1 < 0 ? –rs1 : rs1                                              |
 +-----------------+-------------------------+------------------------------------------------------------------------+
-| **cv.slet**     | **rD, rs1, rs2**        | rD = rs1 <= rs2 ? 1 : 0                                                |
+| **cv.slet**     | **rD, rs1, rs2**        | rD[0] = rs1 <= rs2 ? 1 : 0                                             |
```
davideschiavone, line 605: Hi @pascalgouedo , I do not agree that writing rd[0] is appropriate, it looks like the bits 31:1 are untouched, instead, they are set to 0, see the ALU result:

https://github.com/openhwgroup/cv32e40p/blob/master/rtl/cv32e40p_alu.sv#L973

I suggest we leave it as it was

docs/source/instruction_set_extensions.rst
```diff
@@ -586,10 +586,10 @@ General ALU Operations
...
 |                 |                         | Note: Comparison is signed                                             |
 +-----------------+-------------------------+------------------------------------------------------------------------+
-| **cv.sletu**    | **rD, rs1, rs2**        | rD = rs1 <= rs2 ? 1 : 0                                                |
+| **cv.sletu**    | **rD, rs1, rs2**        | rD[0] = rs1 <= rs2 ? 1 : 0                                             |
```
davideschiavone, line 609: same as above

davideschiavone (changes requested):

docs/source/instruction_set_extensions.rst
pascalgouedo: Hi @davideschiavone 
Yes you are right. I restore previous description for both SLET instructions.

PR #738, opened 2022-11-18
Restored all broken links to external documents or repositories.

pascalgouedo:


docs/source/fpu.rst
```diff
@@ -26,7 +26,7 @@ precision can be enabled by setting the parameter **FPU** of the toplevel file
 The actual Floating Point Unit (FPU) is instantiated outside the
 CV32E40P and is accessed via the APU interface (see :ref:`apu`).
 The FPU repository used by the CV32E40P core is available at
-https://github.com/pulp-platform/fpnew.
+https://github.com/openhwgroup/cvfpu
```
MikeOpenHWGroup, line 29: While it is a good idea to make this change, it is not strickly necessary.  The `pulp-platform` URL will be automatically redirected to the `openhwgroup` URL because the repo was _moved_, not forked.

MikeOpenHWGroup (changes requested):
Recommend using [GitHub permalinks](https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files).

MikeOpenHWGroup (approved):

Issue #452, opened 2020-08-19
Some of the Xpulp instructions use opcodes that are either reserved opcodes or standard opcodes


Silabs-ArjanB:
Some of the Xpulp opcodes have been chosen such that upstreaming of the related tool changes will become impossible unless we change some opcodes. Ideally only 'custom opcodes' are used for Xpulp extensions.

First question to answer:

- [ ] Are there enough custom opcodes  to host all the Xpulp extensions that are now spread in the reserved and standard opcodes?

Craig Blackmore's original message:

As per the actions of the minutes above, here is a summary of opcode issues followed by a breakdown of opcodes used by each CORE-V extension.

The issues fall into 4 categories:

1. Use of standard opcode (e.g. OP, LOAD, BRANCH). Any CORE-V extension that uses these opcodes may conflict with a future RISC-V standard extension.
2. Use of reserved opcode. Any CORE-V extension that uses a reserved opcode may conflict with a future RISC-V standard extension.
3. Use of custom-<n> opcode. The RISC-V Foundation does not allow any standard extension to use the custom opcodes. Therefore, any CORE-V extension that only uses custom opcodes is guaranteed not to conflict with future standard extensions. However, any custom opcodes must be changed before a CORE-V extension can be accepted as a standard extension.
4. Incorrect use of standard opcode. The bitmanip immediate instructions (p.extract, p.extractu, p.insert, p.bclr, p.bset) use OP, but they should use OP-IMM (if we choose to continue using standard opcodes).

Any CORE-V extension that uses standard or reserved opcodes risks conflicting with future standard extensions. Using custom opcodes eliminates that risk, but those opcodes must be changed if the extension is to be standardized. I suspect we would also need a compelling reason to use a reserved opcode rather than fitting into the existing ones. Crucially, whatever encodings we choose now are subject to change during standardization.

Here is a summary of opcodes used by each CORE-V extension:

- PULP bitmanip
   All instructions use OP.
   Immediate instructions (p.extract, p.extractu, p.insert, p.bclr, p.bset) should use OP-IMM.

- PULP SIMD - reserved opcode.

- Post-increment load/store - custom-0 opcode.

- PULP general ALU - Add-shift instructions (p.add[u][R]N[r] and p.sub[u][R]N[r]) use custom-3 opcode.
   Multiply-accumulate instructions use custom-2 opcode.
   Remaining instructions use OP opcode.

- PULP immediate branching - BRANCH opcode.

[...]

PR #740, opened 2022-11-25
Instructions re-encoding aligned with stable pulp_encoding_blocks-OPHW-2022-11-14.xlsx

pascalgouedo:
pulp_encoding_blocks-OPHW-2022-11-14.xlsx from issue https://github.com/openhwgroup/cv32e40p/issues/452.

Changes wrt pulp_encoding_blocks-OPHW-jm-jpb-pgo-2022-04-26.xlsx :
- new encoding for cv.extract, cv.insert, cv.shuffle, cv.pack, cv.cplxmul, cv.subrotmj, cv.add/subb.div2/4/8
- description fix for cv.insert, cv.clb,  cv.bitrev, cv.avgu
- added missing SIMD cv.sdotsp.sc.b
- updates related to Hardware Loops (2 new instructions, CSR mapping, constraints).


Silabs-ArjanB:
Opting out of PULP related reviews

MikeOpenHWGroup (commented):
LGTM, thanks for this @pascalgouedo. I will leave it to @davideschiavone to approve as he is far more knowledgeable than me regarding the XPULP ISA.

docs/source/control_status_registers.rst
```diff
@@ -335,10 +335,31 @@ Detailed:
 | 31:0        | RW        | Number of iteration of HWLoop 0/1.        |
 +-------------+-----------+-------------------------------------------+
 
+.. _csr-uhartid:
+
+User Hardware Thread ID (``uhartid``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+CSR Address: 0xCC8 (only present if ``PULP_XPULP`` = 1)
```
davideschiavone, line 343: similarly as above, do we need this only when PULP SECURE is 1?

docs/source/corev_hw_loop.rst
```diff
@@ -43,76 +43,84 @@ Hardware Loop constraints
...
 The HWLoop constraints are:
 
--  Start and End address of an HWLoop must be word aligned
+-  Start and End address of an HWLoop must be word aligned.
+
+-  End address of an HWLoop must point to the instruction just after the last one of the loop body.
 
 -  HWLoop body must contain at least 3 instructions.
-   An illegal exception is raised otherwise.
+   An illegal instruction exception is raised otherwise.
```
davideschiavone, line 51: when does this happen? when setting the END of the HW Loop? We should specify this in my opinion, as well as specify the ordering of settings, because if you check that the HW loop END instruction is greater or equal to START + 12 (or +16 actually as you point to the next after the end)- when setting the END of the loop, then START must be programmed before, otherwise, if you check when fetching the first time the first instruction of the HW Loop, then ordering does not matter.

docs/source/corev_hw_loop.rst
```diff
@@ -43,76 +43,84 @@ Hardware Loop constraints
...
 
 -  No Compressed instructions (RVC) allowed in the HWLoop body.
-   An illegal exception is raised otherwise.
+   An illegal instruction exception is raised otherwise.
```
davideschiavone, line 54: shall we say this happens when fetching the first time a compressed instruction inside the HW loop?

docs/source/corev_hw_loop.rst
```diff
@@ -43,76 +43,84 @@ Hardware Loop constraints
...
 
--  No uncoditional jump instructions allowed in the HWLoop body.
-   An illegal exception is raised otherwise.
+-  No unconditional jump instructions allowed in the HWLoop body.
+   An illegal instruction exception is raised otherwise.
```
davideschiavone, line 57: shall we say this happens when fetching the first time a jump instruction inside the HW loop?

docs/source/corev_hw_loop.rst
```diff
@@ -43,76 +43,84 @@ Hardware Loop constraints
...
 
--  No coditional branch instructions allowed in the HWLoop body.
-   An illegal exception is raised otherwise.
+-  No conditional branch instructions allowed in the HWLoop body.
+   An illegal instruction exception is raised otherwise.
```
davideschiavone, line 60: shall we say this happens when fetching the first time a branch instruction inside the HW loop?

docs/source/corev_hw_loop.rst
```diff
@@ -43,76 +43,84 @@ Hardware Loop constraints
...
 
 -  No privileged instructions (mret, dret, ecall, wfi) allowed in the HWLoop body, except for ebreak.
-   An illegal exception is raised otherwise.
+   An illegal instruction exception is raised otherwise.
```
davideschiavone, line 63: as above

docs/source/corev_hw_loop.rst
```diff
@@ -43,76 +43,84 @@ Hardware Loop constraints
...
 
 -  No memory ordering instructions (fence, fence.i) allowed in the HWLoop body.
-   An illegal exception is raised otherwise.
+   An illegal instruction exception is raised otherwise.
```
davideschiavone, line 66: as above

docs/source/corev_hw_loop.rst
```diff
@@ -43,76 +43,84 @@ Hardware Loop constraints
...
 
 -  The End address of the outermost HWLoop (#1) must be at least 2
    instructions further than the End address innermost HWLoop (#0),
    i.e. HWLoop[1].endaddress >= HWLoop[0].endaddress + 8
-   An illegal exception is raised otherwise.
+   An illegal instruction exception is raised otherwise.
```
davideschiavone, line 71: as above

docs/source/control_status_registers.rst
```diff
@@ -95,21 +95,21 @@ instruction exception.
...
   +---------------+-------------------+-----------+---------------------+---------------------------------------------------------+
-  | 0x806         | ``lpcount1``      | URW       | ``PULP_XPULP`` = 1  | Hardware Loop 1 Counter.                                |
+  | 0xCC6         | ``lpcount1``      | URO       | ``PULP_XPULP`` = 1  | Hardware Loop 1 Counter.                                |
   +---------------+-------------------+-----------+---------------------+---------------------------------------------------------+
-  | 0xCC0         | ``uhartid``       | URO       | ``PULP_XPULP`` = 1  | Hardware Thread ID                                      |
+  | 0xCC8         | ``uhartid``       | URO       | ``PULP_XPULP`` = 1  | Hardware Thread ID                                      |
   +---------------+-------------------+-----------+---------------------+---------------------------------------------------------+
-  | 0xCC1         | ``privlv``        | URO       | ``PULP_XPULP`` = 1  | Privilege Level                                         |
+  | 0xCC9         | ``privlv``        | URO       | ``PULP_XPULP`` = 1  | Privilege Level                                         |
```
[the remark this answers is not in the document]
pascalgouedo: Agreed.

docs/source/control_status_registers.rst
pascalgouedo: Agreed.

docs/source/corev_hw_loop.rst
pascalgouedo: Same as above.

docs/source/corev_hw_loop.rst
pascalgouedo: Same as above.

docs/source/corev_hw_loop.rst
pascalgouedo: Same as above.

docs/source/corev_hw_loop.rst
pascalgouedo: Same answer as above.

docs/source/corev_hw_loop.rst
pascalgouedo: Quite complex ordering constraint.
A Loop Enable bit could solve the ordering problem when programming HW loops CSRs.
General answer could be what is stated in last conversation/comment.

docs/source/corev_hw_loop.rst
pascalgouedo: Like the "at least 3 instructions", those 2 constraints are harder to tackle in hardware. It would require to add resources (adders/substractors) just to check these wrong cases. Not really nice to keep CV32E40P small and power efficient.

In my previous life, these cases were only managed by constraining compiler and were not checked by HW ...

docs/source/corev_hw_loop.rst
davideschiavone: sorry what is the last conversation/comment?

Anyway, for me it is fine any kind of solution, e.g. the easiest would be when fetching the first instruction of the HW Loop (start), but we need to say this

pascalgouedo: Do nothing in HW for that ...

docs/source/corev_hw_loop.rst
pascalgouedo: None required if no exception.
For SW, they should not generate those cases so no pb.
For Reference Model, emulate an instruction exception.
For verification, those cases should not be verified and so not generated. If it would happen anyway, there will be a difference for sure as Reference Model will generate an exception and the HW will not ...

davideschiavone: mmm I wouldn't go this direction, but I do not have strong opinions either.

We won't definitively sell it as "safe" then, but even if you do nothing in HW, the HW won't work, so we should write AT LEAST that the behaviour is undefined, and we should remove the statements that we raise an exception, as we won't - did I understand this correctly?

davideschiavone: ok then you can say that the reference model does it, but I would not, I mean, in theory, the reference model should behave as the DUT, but I let @MikeOpenHWGroup chime in here

MikeOpenHWGroup: Actually, I always say that the DUT should behave as the reference model.  :wink:

Seriously, this is an important topic.  In the general case there are three conditions:
1. **Illegal events that result in either unpredictable behavior and/or an unrecoverable state.**
The testbench will be constrained to avoid generating these events.  The reference model will not check for these events.

2. **Illegal events that cause an exception.**
The testbench will be able to generate these events and will have functional coverage to ensure that it does.  The reference model will also generate an exception.

3. **Illegal events that are automatically recoverable in the DUT.**
The testbench will be able to generate these events and will have functional coverage to ensure that it does.  The reference model will recover in the same way as the DUT.

pascalgouedo: I generally agree with the first statement **except** when the reference model is not only used just for verification but also as an execution model for software guys when developing their application.
This is the case when this RM is embedded in a Virtual Platform.
If the RM is not generating an exception for case 1. then you are finding the issue only when simulating on RTL or FPGA.
If the compiler is not generating this case no pb but what if "someone" writes assembly code not respecting hw loops constraints?
I know I know writing assembly code !!!

docs/source/corev_hw_loop.rst
```diff
@@ -21,98 +21,105 @@ CORE-V Hardware Loop Extensions
...
-   i.e. HWLoop[1].endaddress >= HWLoop[0].endaddress + 8
-   An illegal exception is raised otherwise.
+The rationale of putting the end-of-loop label to the first instruction after the last one of the loop body
+is that it greatly simplifies compiler optimization (relative to basic blocks management).
 
-In order to use hardware loops, the compiler needs to setup the loop
-beforehand with the following instructions. Note that the minimum loop
-size is 3 instructions and the last instruction cannot be any jump or
-branch instruction.
+In order to use hardware loops, the compiler needs to setup the loops beforehand with the 6 defined instructions.
```
davideschiavone, line 79: In order to use hardware loops, the compiler needs to setup the loops beforehand with the cv.start/i, cv.count/i, and cv.stop/i instructions.

davideschiavone (changes requested):

PR #760, opened 2023-01-26
Update documentation of instruction extensions for CV32E40P version 2.

jeremybennett:
These changes particularly relate to how tool chain support is
	invoked using -march.  We take the opportunity to remove caveats
	about ISA extensions not supported in the tool chain, since all
	are now supported in both GCC and Clang/LLVM.

	* docs/source/instruction_set_extensions.rst: Add -march tool chain invocations for each ISA extension and remove all caveats about compiler support.

Signed-off-by: Jeremy Bennett <jeremy.bennett@embecosm.com>


docs/source/instruction_set_extensions.rst
```diff
@@ -20,17 +20,22 @@
...
 
-CV32E40P supports the following CORE-V ISA Extensions, which are part of **Xcorev** and can be enabled by setting ``PULP_XPULP`` == 1.
+CV32E40P supports the following CORE-V ISA Extensions, which can be enabled by setting ``PULP_XPULP`` == 1.
 
- * Post-Incrementing load and stores, see :ref:`corev_load_store`.
- * Hardware Loop extension, see :ref:`corev_hardware_loop`.
- * ALU extensions, see :ref:`corev_alu`.
- * Multiply-Accumulate extensions, see :ref:`corev_multiply_accumulate`.
- * Single Instruction Multiple Data (aka SIMD) extensions, see :ref:`corev_simd`.
+ * Post-Incrementing load and stores, see :ref:`corev_load_store`, invoked in the tool chain with ``-march=xcvmem``.
```
pascalgouedo, line 25: Do we need to be more precise like ``-march=rv32i*_xcvmem``

docs/source/instruction_set_extensions.rst
```diff
@@ -20,17 +20,22 @@
...
+   * miscellaneous ALU instructions, invoked in the tool chain with ``-march=xcvalu``; and
+   * immediate branch instructions, invoked in the tool chain with ``-march=xcvbi``.
 
-To use such instructions, you need to compile your SW with the CORE-V GCC compiler.
+ * Multiply-Accumulate extensions, see :ref:`corev_multiply_accumulate`, invoked in the tool chain with ``-march=xcvmac``.
+ * Single Instruction Multiple Data (aka SIMD) extensions, see :ref:`corev_simd`, invoked in the tool chain with ``-march=xcvsimd``.
+
+Additionally the event load instruction (**cv.elw**) is supported by setting ``PULP_CLUSTER`` == 1, see :ref:`corev_event_load`.  This is a separate ISA extension, invoked in the tool chain with ``-march=xcvelw``.
+
+To use such instructions, you need to compile your SW with the CORE-V GCC or Clang/LLVM compiler.
```
pascalgouedo, line 38: Maybe we should specify starting from which version those instructions are supported by the toolchain, GCC and LLVM.

docs/source/instruction_set_extensions.rst
```diff
@@ -1066,9 +1071,6 @@ performed.
 
 The custom SIMD extensions are only supported if ``PULP_XPULP`` == 1.
 
-**SIMD is not supported by the compiler of the tool chain** as it is not implementing auto-vectorization up to now.
-But those instructions can be used either with builtins or even in assembly.
-
```
pascalgouedo, line 1071: I think we should keep this comment because it is still true.

docs/source/instruction_set_extensions.rst
jeremybennett: That's a good suggestion. I'll make this change.

docs/source/instruction_set_extensions.rst
jeremybennett: A good suggestion. I'll make this change.

docs/source/instruction_set_extensions.rst
jeremybennett: I will modify it. SIMD will be supported, but we won't have autovectization.

jeremybennett:
All comments by @pascalgouedo accepted and updated version force pushed.

davideschiavone:
@pascalgouedo can you do the merge as you made the review?

davideschiavone (approved):

PR #799, opened 2023-04-20
Adding FPU & ZFINX support on rvfi

YoannPruvost:


bhv/cv32e40p_rvfi_trace.sv
```diff
@@ -75,14 +87,31 @@ module cv32e40p_rvfi_trace
   logic [31:0] imm_shuffle_type;
   logic [ 4:0] imm_clip_type;
 
-  assign rd = rvfi_rd_addr[0];
-  assign rs1 = rvfi_rs1_addr;
-  assign rs2 = rvfi_rs2_addr;
+  always_comb begin
+    if (rvfi_frs1_rvalid) begin
+      rs1 = 6'b100000 | rvfi_frs1_addr;
```
davideschiavone, line 92: using the OR makes it less readable, can we just extend it?

like

```
rs1 = {1'b1, rvfi_frs1_addr};
```

bhv/cv32e40p_rvfi_trace.sv
```diff
@@ -112,7 +141,7 @@ module cv32e40p_rvfi_trace
       .is_compressed_o(is_compressed)
   );
 
-  localparam FPU = 0;
+  localparam FPU = 1;
```
davideschiavone, line 144: should we have this as a parameter?

bhv/pipe_freeze.sv
```diff
@@ -0,0 +1,632 @@
...
+
+// CV32E40P
+//
+// Contributors: Yoann Pruvost, Dolphin Design <yoann.pruvost@dolphin.fr>
+
+/*
+   * This struct is used to store all information comming from the core at every posedge
+   * The information will then be processed
+   */
+typedef struct {
```
davideschiavone, line 25: if this module is used only in the tracer, can you please call it pipe_freeze_trace.sv?

davideschiavone (changes requested):

bhv/pipe_freeze.sv
YoannPruvost: done

bhv/cv32e40p_rvfi_trace.sv
YoannPruvost: you are right
it's done ;)

bhv/cv32e40p_rvfi_trace.sv
YoannPruvost: We need them to reuse the log file part of the former tracer.
Still I moved it from localparam to global param to match the core configuration

PR #812, opened 2023-05-24
HW loop csr + improved FPU instruction 

YoannPruvost:
This pull request add two 

- HW loop support on rvfi
- Improved floating point support. Instructions using the fpu are now issued to a dedicated fifo allowing standard instructions to be monitored at the same time.


davideschiavone:
replaced by #819

PR #823, opened 2023-06-23
User Manual updates

pascalgouedo:
- Added a constraint about HWloop start, and and setup instructions alignment.
- Corrected assembly example wrt assembly alignment directives.
- Changed parameters values for Zfinx CSR presence.


MikeOpenHWGroup (approved):
These are documentation-only updates, so I am approving/merging, irregardless of the CI EqChk.

Issue #840, opened 2023-07-31
fix mimpid 


davideschiavone:
# Fix `mimpid` 

According to https://docs.openhwgroup.org/projects/cv32e40p-user-manual/en/latest/core_versions.html#cv32e40p-v2-0-0, we need to set the `mimpid` to 1 when the parameters related to custom and floating-point ISA extensions are set to inform the users that the core is not logically equivalent to the frozen version v1.0.0.

davideschiavone:
The solution is trivial and it is about changing this line https://github.com/openhwgroup/cv32e40p/blob/master/rtl/cv32e40p_cs_registers.sv#L509 to split between MIMPID and MTVAL values, and return 1 in the MIMPID whenever one of the above parameters is set.

Issue #838, opened 2023-07-25
pulp hwloop count not decrementing to 0 after the end of hwloop body execution


dd-vaibhavjain:
### Issue Description
In pulp hwloop tests, after the end of hwloop body, i.e., after the last hwloop body instruction has been executed, the loop count csr does not decrement to 0 and final value of count remains at 1

### Component
**Component:RTL**

### Steps to Reproduce

1. cv32e40p git hash : 99772238edb4093e99386599b9f3d79f4d125f30

2.  Using "core-v-verif" repo ,  branch : cv32e40p/dev
     core-v-verif hash :  c7610d8ab6fecdb81b21a6e4e791c8f9e01fc7bd
     
-      git clone -b cv32e40p/dev https://github.com/openhwgroup/core-v-verif.git
-      cd core-v-verif
-      git checkout c7610d8ab6fecdb81b21a6e4e791c8f9e01fc7bd
  
     
3. Run Test:  
Run with reference model (USE_ISS=yes) to get the error mismatch in log. Otherwise if running without USE_ISS, we can see this issue from the wave dump.
-     setenv SIMULATOR vsim (or any other simulator)
-     cd cv32e40p/sim/uvmt/
-      make test TEST=pulp_hardware_loop USE_ISS=yes CFG=pulp_fpu WAVES=1 SEED=1 RUN_INDEX=1 USER_RUN_FLAGS="+IDV_TRACE2LOG=1"

4. Log Error:

> Info (IDV) Instruction executed prior to mismatch '0x1c0(startZ_1+8): 00188893 addi    x17,x17,1'
> Error (IDV) CSR register value mismatch (HartId:0, PC:0x000001c0 startZ_1+8):
> Info (IDV)  0> CSR cc2 (lpcount0)
> Info (IDV)   . dut:0x00000001
> Info (IDV)   . ref:0x0000000

davideschiavone:
can we close this as now 788caae7d5e4d5fc07f68dc7f34f6307ca17f5d4 has been merged?

dd-vaibhavjain:
Thanks @pascalgouedo for this fix. 
Verified the RTL fix and it resolves the issue here. RTL and ISS model both are aligned for hwloop tests. 
Closing this ticket.

Issue #730, opened 2022-09-27
Custom Xpulp instructions update register file wrongly | Forwarding issue


salaheddinhetalani:
### Issue Description
Custom XPULP instructions, CV.ADDxNR and CV.SUBxNR, read invalid source operand in case they are preceded by a multicycle F instruction and their source register is the same destination register of the floating point one, thus resulting wrong register file update.

### Component
**Component:RTL**

### RISC-V Specification
> The Zfinx extension adds all of the instructions that the F extension adds, except for the transfer instructions FLW, FSW, FMV.W.X, FMV.X.W, C.FLW[SP], and C.FSW[SP].

>The Zfinx variants of these F-extension instructions have the same semantics, except that whenever such an instruction would have accessed an f register, it instead accesses the x register with the same number.

### Steps to Reproduce

As shown below, the following sequence of instructions happens: 
> fdiv.s x5, x21, x4, RTZ -> cv.addunr x13, x1, x5 

The instruction `cv.addunr` is decoded at `t##0` and executed updating the integer register file at `t##1` while setting the wrong result of `32'h189` into `x13`. This is due to reading the old value `32'h20e57` of `x5` instead of the new one of `32'h0` updated at `t##2` by the previously fetched `fdiv.s`. 

![Bug_26](https://user-images.githubusercontent.com/28307467/192526217-f62bd258-ae04-41f0-9d93-5ab0213407f9.png)

**Top Level Parameters**

```
cv32e40p_wrapper #(
    .PULP_XPULP       (1),
    .PULP_CLUSTER     (1),
    .FPU              (1),
    .PULP_ZFINX       (1),
    .NUM_MHPMCOUNTERS (1)
)
```
Git Hash: TBU
Flist: [cv32e40p_fpu_manifest.flist](https://github.com/openhwgroup/cv32e40p/blob/master/cv32e40p_fpu_manifest.flist.) 
VCD: [bug_26.vcd](https://github.com/openhwgroup/cv32e40p/files/9657041/bug_26.zip)

---
Product: OneSpin 360 DV-Verify
App: Processor Verification App
Tool's version: 2022.3_1

davideschiavone:
can we close this as this one is now merged? 6cfb7438ba62fb433823c85d04e2e001d56e5b81

pascalgouedo:
Work in on-going on bug resolution w/ formal verification guy.
We will close all resolved opened issue together once confirmed by formal runs.

[...]

PR #841, opened 2023-07-31
3 RTL Issues corrections

pascalgouedo:
Issue #838 correction
Issue #730 correction
Issue #840 correction

### DIFF
--- a/rtl/cv32e40p_cs_registers.sv
+++ b/rtl/cv32e40p_cs_registers.sv
@@ -507,7 +507,7 @@ module cv32e40p_cs_registers
 
         // mimpid, Machine Implementation ID
         CSR_MIMPID: begin
-          csr_rdata_int = (FPU || COREV_PULP || COREV_CLUSTER || (NUM_MHPMCOUNTERS > 1)) ? 32'h1 : 'b0;
+          csr_rdata_int = (FPU || COREV_PULP || COREV_CLUSTER) ? 32'h1 : 'b0;
         end
 
         // unimplemented, read 0 CSRs
--- a/rtl/cv32e40p_cs_registers.sv
+++ b/rtl/cv32e40p_cs_registers.sv
@@ -505,8 +505,11 @@ module cv32e40p_cs_registers
         // marchid: Machine Architecture ID
         CSR_MARCHID: csr_rdata_int = MARCHID;
 
+        // mimpid, Machine Implementation ID
+        CSR_MIMPID: csr_rdata_int = (FPU || COREV_PULP || COREV_CLUSTER || (NUM_MHPMCOUNTERS > 1)) ? 32'h1 : 'b0;
+
         // unimplemented, read 0 CSRs
-        CSR_MIMPID, CSR_MTVAL: csr_rdata_int = 'b0;
+        CSR_MTVAL: csr_rdata_int = 'b0;
 
         CSR_TSELECT, CSR_TDATA3, CSR_MCONTEXT, CSR_SCONTEXT: csr_rdata_int = 'b0;  // Always read 0
         CSR_TDATA1: csr_rdata_int = tmatch_control_rdata;
--- a/rtl/cv32e40p_cs_registers.sv
+++ b/rtl/cv32e40p_cs_registers.sv
@@ -506,7 +506,9 @@ module cv32e40p_cs_registers
         CSR_MARCHID: csr_rdata_int = MARCHID;
 
         // mimpid, Machine Implementation ID
-        CSR_MIMPID: csr_rdata_int = (FPU || COREV_PULP || COREV_CLUSTER || (NUM_MHPMCOUNTERS > 1)) ? 32'h1 : 'b0;
+        CSR_MIMPID: begin
+          csr_rdata_int = (FPU || COREV_PULP || COREV_CLUSTER || (NUM_MHPMCOUNTERS > 1)) ? 32'h1 : 'b0;
+        end
 
         // unimplemented, read 0 CSRs
         CSR_MTVAL: csr_rdata_int = 'b0;
--- a/rtl/cv32e40p_controller.sv
+++ b/rtl/cv32e40p_controller.sv
@@ -223,6 +223,8 @@ module cv32e40p_controller import cv32e40p_pkg::*;
   logic hwlp_end1_eq_pc;
   logic hwlp_counter0_gt_1;
   logic hwlp_counter1_gt_1;
+  logic hwlp_counter0_eq_1;
+  logic hwlp_counter1_eq_1;
   logic hwlp_end0_eq_pc_plus4;
   logic hwlp_end1_eq_pc_plus4;
   logic hwlp_start0_leq_pc;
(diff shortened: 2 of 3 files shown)


rtl/cv32e40p_cs_registers.sv
```diff
@@ -505,8 +505,13 @@ module cv32e40p_cs_registers
         // marchid: Machine Architecture ID
         CSR_MARCHID: csr_rdata_int = MARCHID;
 
+        // mimpid, Machine Implementation ID
+        CSR_MIMPID: begin
+          csr_rdata_int = (FPU || COREV_PULP || COREV_CLUSTER || (NUM_MHPMCOUNTERS > 1)) ? 32'h1 : 'b0;
```
davideschiavone, line 510: why are you explicitly checking the NUM_MPHCOUNTERS? this parameter was already part of v1.0.0

davideschiavone (changes requested):

rtl/cv32e40p_cs_registers.sv
davideschiavone: to me looks strange, are we verifying the v2 for each possible number of counters?
If not, what are the differences (and why) between v1 and v2 when we have more than 1 counter?

pascalgouedo: not all possible counters number but maybe going up to 2 or 4 for v2 ...

davideschiavone: are v1 and v2 LEC when the number of counters are 4? and if not, why?

pascalgouedo: v1 is using default value 1 for NUM_MHPMCOUNTERS.
if we force 4 in v1 golden ref and in v2 they are LEC.
but as v1 verification didn't verify anything else than value 1, how to be sure the 3 new counters are working properly?
=> going to be verified in v2 config only (maybe, just thinking about that).
we are not re-opening v1 verif so value 1 is mandatory.
you are thinking about if no bug found in v2 and value = 4, then we can authorize v1 with value = 4 (as long as LEC)?

pascalgouedo: all 29 mhpmcounter CSR accesses are checked in mhpmcounter29_csr_access_test_* tests.
but all event sources are only checked on mhpmcounter3 in perf_counters_instructions test.

davideschiavone: but if v1 and v2 are LEC in the number of counters (up to 4 at least), does it matter that none verified that v1 worked with 4 counters if it is equivalent to v2 - that is my 2 cents 

If you verified v2 with 7 counters, and I do a project with 8 counters and tag it as v3, shall I change the mimpid even if they are LEC? I would not do it...

@Silabs-ArjanB , what do you think?

pascalgouedo: > If you verified v2 with 7 counters, and I do a project with 8 counters and tag it as v3, shall I change the mimpid even if they are LEC? I would not do it...

This is a point I though about as well when updating the User Manual.
If we want any one to be free to use any value for NUM_MHPMCOUNTERS then we must verify that all 29 counters are working properly.

@MikeOpenHWGroup Any thought on that point?

pascalgouedo: > but if v1 and v2 are LEC in the number of counters (up to 4 at least), does it matter that none verified that v1 worked with 4 counters if it is equivalent to v2 - that is my 2 cents

So you rely on v2 verification to ensure v1 + NUM_MHPMCOUNTERS > 1 is OK

Silabs-ArjanB: > but if v1 and v2 are LEC in the number of counters (up to 4 at least), does it matter that none verified that v1 worked with 4 counters if it is equivalent to v2 - that is my 2 cents
> 
> If you verified v2 with 7 counters, and I do a project with 8 counters and tag it as v3, shall I change the mimpid even if they are LEC? I would not do it...
> 
> @Silabs-ArjanB , what do you think?

I would also not change mimpid if there is no actual RTL change (but just more extensive verification)

MikeOpenHWGroup:
> This is a point I though about as well when updating the User Manual.
If we want any one to be free to use any value for NUM_MHPMCOUNTERS then we must verify that all 29 counters are working properly.

Yes, that is my opinion also.  It is fine to say "only NUM_MHPMCOUNTERS == 1 has been verified" in the User Manual.  By if we allow the parameter to be set to any value, then we need to verify that is works.  This particular feature is high-effort, low-reward, so I recommend we stick to num-counters==1.  If a user of our IP wants to use a different number of counters, they can verify that for themselves.

pascalgouedo:
So
- v1 : mimpid = 0 implies NUM_MHPMCOUNTERS == 1
- v2 : mimpid = 1 implies NUM_MHPMCOUNTERS == 1
- NUM_MHPMCOUNTERS > 1 : mimpid = 2?
or totally remove NUM_MHPMCOUNTERS check for mimpid?

MikeOpenHWGroup:
Hmm.  We should not remove the NUM_MHPMCOUNTERS check for mimpid since increasing the number of counters is clearly a software-impacting change.  I would suggest `NUM_MHPMCOUNTERS > 1 : mimpid >= 2`.  Keep in mind that to have a globally unique identifier for the core you need at least `mimpid`, `marchid` and `mvendorid`.

Silabs-ArjanB:
> or totally remove NUM_MHPMCOUNTERS check for mimpid?

As long as no actual RTL changes are made related to NUM_MHPMCOUNTERS (more specifically if the RTL is LEC clean for the original parameter set and SEC clean for any other NUM_MHPMCOUNTERS value that you want to claim to support) , then _mimpid shall not be changed either_ because of a change in verified NUM_MHPMCOUNTERS value(s).

davideschiavone:
> > or totally remove NUM_MHPMCOUNTERS check for mimpid?
> 
> As long as no actual RTL changes are made related to NUM_MHPMCOUNTERS (more specifically if the RTL is LEC clean for the original parameter set and SEC clean for any other NUM_MHPMCOUNTERS value that you want to claim to support) , then _mimpid shall not be changed either_ because of a change in verified NUM_MHPMCOUNTERS value(s).

completely agree - LEC is the strongest point of differentiating between the 2 versions - 

so if v2 and v1 are LEC on the number of counters, then why one would change the mimpid? 


The mimpid is used by the software - would the software change between v1 with 4 counters and v2 with 4 counters (with all the other parameters set to their v1 version?) if the answer is no, which I understood it is as the 2 are LEC, then there is no need for changing the mimpid. 

So for this specific PR, I would simply remove the counters from the OR equation on the mimpid CSR assignment.

pascalgouedo:
LEC between v1 and v2 with NUM_MHPMCOUNTERS = 4 doesn't mean the 3 counters not verified neither for v1 and v2 are working fine.
For me mimpid = 0 or 1 means "verified v1 & v2 version with its defined set of parameters".
Any other value means "not verified set of parameters" as of today.
This is not used to understand how many counters you have, you can figure out that with some smart C function using CSR accesses and illegal exceptions.

davideschiavone:
for me v1 and v2 means they are not LEC and thus the SW needs to differentiate. It is not an ID of the verification but it serves SW. 

If by any chance the verification v2 would end up be identical to v1 even on the FPU or PULP, then we would not change the mimip, even if they were unverified.

Silabs-ArjanB:
@pascalgouedo The sentence you quote is there indeed, but I consider it inaccurate. Also it conflicts with the remaining content of that chapter (as well as our intent)

davideschiavone:
Agree with @Silabs-ArjanB , the sentence is not accurate - so what I vote for is to remove the NUM_COUNTERS from the `mimpid` csr assignment. We can add in the documentation up to which number they have been verified and whether they are LEC up to that number

PR #853, opened 2023-08-18
Get trap information directly when instruction complete ID

YoannPruvost:
Previous implementation of trap detection was recreating minstret counter and using it to check trap. This one directly takes the information from ID.

### DIFF
--- a/bhv/cv32e40p_rvfi.sv
+++ b/bhv/cv32e40p_rvfi.sv
@@ -1190,7 +1190,6 @@ module cv32e40p_rvfi
 
     s_test_for_dret = 1'b0;
 
-    $display("*****Starting pipeline computing*****\n");
     forever begin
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup:
I know they have been there for sometime already but I would recommend either removing the `$display()` statements altogether or prefixing with "%m" and "%0t" so that the source of the message can be easily determined.  Something like this:
```
$display("%m @ %0t: **** starting pipeline ****", $time);
```

YoannPruvost:
> I know they have been there for sometime already but I would recommend either removing the `$display()` statements altogether or prefixing with "%m" and "%0t" so that the source of the message can be easily determined. Something like this:
> 
> ```
> $display("%m @ %0t: **** starting pipeline ****", $time);
> ```

Agree.
I removed it

davideschiavone (approved):

PR #856, opened 2023-08-23
RVFI updates for pulp post increment load instr, all CSR updates for FPU instructions and some general improvements

dd-vaibhavjain:
This PR is for updaing RVFI for issues seen during simulations

1.  Fix for post-increment loads - use trace * m_rd_addr[0], m_rd_wdata[0] ,i.e, index 0 for any reg file updates from ALU for instructions with multiple register updates such as post increment loads.
2. Clean trace_wb to only move load instructions to this wb trace structure.
3. Fix FCSR updates to ensure this register update are captured for both APU instructions and explicit CSR updates. And also update fflags and frm CSRs accordingly.
4. Add logic to check for mstatus FS and SD fields update from FPU instructions.
5. Add INIT_CSR macro and corresponding function init_csr() for insn_trace to ensure each new object has a clean CSR values. Thus to ensure clean CSR update status for each instruction and also simplify debugging for CSR update issues for RVFI and reference model.
6. Check got_mistret flag is set or not before updating mistret csr from EX stage to avoid multiple updates to the CSR from ID and EX stage.

### DIFF
--- a/bhv/cv32e40p_rvfi.sv
+++ b/bhv/cv32e40p_rvfi.sv
@@ -135,6 +135,7 @@ module cv32e40p_rvfi
     input logic        rf_we_wb_i,
     input logic [ 5:0] rf_addr_wb_i,
     input logic [31:0] rf_wdata_wb_i,
+    input logic        regfile_alu_we_ex_i,
     // LSU
     input logic [31:0] lsu_rdata_wb_i,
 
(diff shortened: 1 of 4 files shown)


MikeOpenHWGroup (commented):
This LGTM, but I am not sufficiently familar with the implementation of the RFVI Tracer to offer a valid technical opinion about the changes.  I will request a review from @davideschiavone.

YoannPruvost (approved):
All good to me!

PR #863, opened 2023-08-31
Fix Luimm5 and ls[23] typos

craigblackmore:
These should be Iuimm5 and Is[23].


pascalgouedo:
Feedback taken into account in PR #864.

PR #867, opened 2023-09-06
Better mstatus when FPU instruction

YoannPruvost:
When a floating point instruction cause an update of the mstatus csr, this update was not visble on the rvfi trace for non fpu instruction that were completed before the fpu response but are considered retired after.

This workaround updates the mstatus read value for all instructions that are completed before the fpu response but come latter in retire order

### DIFF
--- a/bhv/cv32e40p_rvfi.sv
+++ b/bhv/cv32e40p_rvfi.sv
@@ -627,7 +627,7 @@ insn_trace_t trace_if, trace_id, trace_ex, trace_ex_next, trace_wb;
   function void empty_fifo();
     integer i, trace_q_size;
     trace_q_size = wb_bypass_trace_q.size();
-    if (trace_q_size > 40) begin
+    if (trace_q_size > 50) begin
       $fatal("Reorder queue too long\n");
     end
     if (trace_q_size != 0) begin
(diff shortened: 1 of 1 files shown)


bhv/cv32e40p_rvfi.sv
```diff
@@ -706,11 +706,11 @@ insn_trace_t trace_if, trace_id, trace_ex, trace_ex_next, trace_wb;
       new_rvfi_trace.m_csr.fflags_rdata = s_fflags_mirror;
       new_rvfi_trace.m_csr.frm_rdata = s_frm_mirror;
       new_rvfi_trace.m_csr.fcsr_rdata = s_fcsr_mirror;
-      if (s_mstatus_sd_fs_mirror != 32'h0) begin
-        new_rvfi_trace.m_csr.mstatus_wdata = new_rvfi_trace.m_csr.mstatus_wdata | s_mstatus_sd_fs_mirror;
-        new_rvfi_trace.m_csr.mstatus_wmask = 32'hFFFF_FFFF;
-        s_mstatus_sd_fs_mirror = 32'h0;  // Reset mirror
-      end
+      // if (s_mstatus_sd_fs_mirror != 32'h0) begin
```
davideschiavone, line 709: can you delete this piece of code instead of commenting it? you can always find it back in the github history

davideschiavone (changes requested):

bhv/cv32e40p_rvfi.sv
YoannPruvost: It's done

PR #929, opened 2024-01-02
Refactor of LEC/SEC scripts

MarioOpenHWGroup:
- LEC script refactor
- SEC process always supported using Siemens SLEC App


pascalgouedo (changes requested):
Did you add those slec scripts to do the same than lec, meaning v1 versus v2 with v1 parameters setting?
Because as of today as the top level is cv32e40p_core and not cv32e40p_top and so it is excluding FPU.
We could extend the methodology for the future to be able to check v2 with v2+++ by moving top_module to cv32e40p_top.
I have done that for lint so I can lint v1 and all configurations of v2...
In fact to deal with configurability, I added a new top module (cv32e40p_wrapper which instantiates cv32e40p_top) which imports a config package which is selected by shell script.

scripts/slec/siemens/Makefile
```diff
@@ -0,0 +1,41 @@
+##############################################################################
+#             Copyright 2006-ntor Graphics Corporation
```
pascalgouedo, line 2: Mentor Graphics

pascalgouedo (changes requested):

MarioOpenHWGroup:
I have also added the Cadence SEC script using JasperGold, if you can review the PR again :)

pascalgouedo (approved):

PR #931, opened 2024-01-08
add RTL LINT scripts.

pascalgouedo:


scripts/lint/cv32e40p_wrapper.sv
```diff
@@ -0,0 +1,113 @@
+// Copyright 2018 ETH Zurich and University of Bologna.
```
davideschiavone, line 1: add OpenHW Group license please

davideschiavone (changes requested):

scripts/lint/cv32e40p_wrapper.sv
pascalgouedo: This is a general topic I wanted to discuss with @MikeOpenHWGroup and you.
None of the RTL files have OpenHW group license, only BHV ones have it.
And RTL headers are not coherent between themselves.
I think we should discuss about that, decide about a rule and create a specific PR just for that subject.

I am going to use this file to propose something.
For each file, we can add all contributors already listed in and keep original file description.

MikeOpenHWGroup: Yes @pascalgouedo, you are right that we are not consistent in the way we manage license header files. It is my understanding that for any file that are substantially modified, a new copyright _may_ be added, but the existing copyright and license header _must not_ be removed or modified. 

Usually I will do this with a one-line SPDX header.  For example:
```
// Copyright 2024 OpenHW Group and <member co.>
// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
// Copyright 2018 ETH Zurich and University of Bologna.
// ...original license header from ETHZ and UniBo.
```
I have created pull-request #933 and we can discuss this idea there (without blocking this PR).

davideschiavone (approved):

PR #934, opened 2024-01-08
Add an example for updating copyright and license header

MikeOpenHWGroup:
As per discussion in pull-request #931.


davideschiavone (approved):

pascalgouedo (commented):
Depending of the files, this is creating a sandwich of different layers which makes it awful to see !!!

All files have the same SHL license. If for each file we keep all previous information but restructured in a common way, I think this should not be a pb.

Some of them have Silabs, some of them ETH, others Bologna only.
I would propose ETH, Bologna, Silabs, Dolphin Design and OpenHW for all RTL files.
Plus keeping already listed contributors file per file.

MikeOpenHWGroup:
> Depending of the files, this is creating a sandwich of different layers which makes it awful to see !!!

Yes, I agree.  I have seen some truly ugly open-source license headers!  This is why I suggest adding only two lines, one for the copyright and a second for the SPDX license.  That adds only two lines per contributor, and if the license does not change then only one line.

pascalgouedo:
Doesn't really answer to the original point of cleaning up...
But if we don't/can't change original header I will add our own layer as Arjan did ...

MikeOpenHWGroup:
> Doesn't really answer to the original point of cleaning up...

I presume you mean this point:
>> None of the RTL files have OpenHW group license, only BHV ones have it.

Agree and understood.  If we agree on the the method for updating copyright and license headers, the next step is to update all the RTL.  I leave this to you, @davideschiavone and maybe @Silabs-ArjanB.

pascalgouedo:
I will propose through a specific PR what I have in mind for all RTL files.
I'll put @davideschiavone, @Silabs-ArjanB and yourself as reviewers.

DBees:
@MikeOpenHWGroup @pascalgouedo @davideschiavone the difficulty I have is that it is very hard to see who is contributing what with multiple authors presented in the same header. In other words, member A contributes something under License 1. Member B then contributes something additional under License 2. Not sure, but I think they should have separate headers with a description of what has been added. Otherwise it is impossible to see who is asserting what rights and granting what licenses to copyrights and patents. Let me check with Eclipse guys.

MikeOpenHWGroup:
> Not sure, but I think they should have separate headers with a description of what has been added.

My proposal already adds a legal and complete license header (that this the comment line starting with SPDX).  The description of what has been added is taken care of by the git/GitHub revision history and record of pull-request. No need to repeat that in a comment header.

pascalgouedo:
I think that from a license point of view, files should be self-sufficient without relying on any file management history.
I have seen somewhere that for every file modification there should have some description additionally to your new copyright/contributors lines.
For sure not too detailed but maybe only related v1 and v2 differences in this present case?
I agree that differences between RI5CY and v1 would be missing.

PR #952, opened 2024-02-28
RTL correction for verilator (no added/removed lines).

pascalgouedo:


pascalgouedo:
Merging it as I need it to progress on X-heep side.

PR #1055, opened 2026-03-31
Update Read the Docs OS version to Ubuntu 24.04

MikeOpenHWGroup:
Read The Docs announced the deprecation of the Ubuntu 20.04 LTS build image (build.os: ubuntu-20.04).  This PR updates to 24.04.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -5,7 +5,7 @@
 version: 2
 
 build:
-  os: "ubuntu-20.04"
+  os: "ubuntu-24.04"
   tools:
     python: "3.9"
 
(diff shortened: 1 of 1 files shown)


cairo-caplan:
Hey @MikeOpenHWGroup , WDYT about making it `ubuntu-lts-latest` instead https://docs.readthedocs.com/platform/stable/config-file/v2.html#build-os ?

It is still a lts version, so issues are not usually expected. At the same time, by being the latest, we wouldn't need to change this in the future ever again.

MikeOpenHWGroup:
>  WDYT about making it ubuntu-lts-latest instead

Good idea.  I am not 100% confident about it, but it is worth a try as it will reduce our maintenance costs.  Let's try it for this repo and see how it goes.  I've pushed in the change.

PR #1056, opened 2026-04-10
Fix FPU simulation issue in Vivado simulator

eimbens:
Problem:

When simulating the core with FPU enabled in Vivado, executing floating point instructions produces incorrect results. 

I traced back the problem to rtl/vendor/pulp_platform_fpnew/src/fpnew_pkg.sv, where an access to any field in the FP_ENCODINGS struct returns 7. More precisely, an access to any field always returns the value stored in the FP_ENCODINGS[FP16ALT].man field, which holds 7. The FP_ENCODINGS struct is used to retrieve the exponent and mantissa sizes in different floating point representations. 

I found the problem while I was creating a block design to run the cv32e40p core with FPU on the PYNQ-Z2 FPGA. While testing the FPU using a Verilog testbench and Vivado simulation, I noticed that the floating point values in the waveforms were incorrect. 

Fix:

To fix the problem I split up the FP_ENCODINGS struct into an FP_EXP and an FP_MAN struct, which store the exponent and mantissa sizes of all representations separately. 

I do not know the root cause of the problem, but I tested the fix in Vivado 2024.1 and 2025.2, and in both cases it fixes the issue without affecting other behavior.

### DIFF
--- a/rtl/vendor/pulp_platform_fpnew/src/fpnew_pkg.sv
+++ b/rtl/vendor/pulp_platform_fpnew/src/fpnew_pkg.sv
@@ -56,6 +56,24 @@ package fpnew_pkg;
     // add new formats here
   };
 
+  localparam int unsigned FP_EXP [0:NUM_FP_FORMATS-1]  = '{
+    8,  // IEEE binary32 (single)
+    11, // IEEE binary64 (double)
+    5,  // IEEE binary16 (half)
+    5,  // custom binary8
+    8   // custom binary16alt
+    // add new formats here
+  };
+
+  localparam int unsigned FP_MANTISSA [0:NUM_FP_FORMATS-1]  = '{
+    23, // IEEE binary32 (single)
+    52, // IEEE binary64 (double)
+    10, // IEEE binary16 (half)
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup:
Hi @eimbens, thanks for your interest in OpenHW and the CV32E40P.  We cannot accept this PR for two reasons:

1. You have not signed the Eclipse Contributor Agreement. In fairness to you, this requirement is not properly documented for this repository.  Fortunately, if you choose to do so, signing is painless, see [CONTRIBUTING](https://github.com/openhwgroup/cv32e20-dv/blob/main/CONTRIBUTING.md#contributor-agreement) in our `cv32e20-dv` repo. (I will update CONTRIBUTING for this repo shorty.)

2. We do not support the Vivado simulator.  The code in question is legal SystemVerilog syntax which simulates and synthesizes correctly using SystemVerilog tools from all of the "big 3" EDA vendors.  The code in this repo is very stable and extensively verified and we will only accept PRs for functional or synthesis bugs or new features that have been approved by the Project Leads.  I suggest that you open a support issue with your tool vendor.

PR #1061, opened 2026-05-29
fix: Replace the blocking assignment in the always_latch block with a non-blocking assignment

nikgiu:
Fix: Replace the blocking assignment in the always_latch block with a non-blocking assignment to avoid mixing blocking and non-blocking assignments in the same procedural block in cv32e40p_register_file_latch.sv.


davideschiavone:
thanks for the fix @nikgiu - can you please sign the ECA? it is needed to contribute to the core

Also, PRs need to be done in the dev branch

nikgiu:
Hello @davideschiavone, sorry for the wrong procedure.
Let me know if everything is in order now.

PR #1065, opened 2026-07-01
Fix #1060

### LINKED ISSUE #1060: mstatus.FS not visible to immediately following csrrs mstatus after flw
## Description
After an flw instruction, mstatus.FS is not yet updated to Dirty when a csrrs mstatus immediately follows. The read returns the pre-flw FS value. One NOP between the two instructions is sufficient to observe the correct (Dirty) value.

This was found  in an ACT4 rv32imcf certification run. 

## Cause

For two consecutive instructions flw and csrrs: At cycle T+1, flw is in its writeback stage and csrrs is in its execute stage at the same time. The writeback sets mstatus_fs_n to Dirty combinatorially within that cycle, but the CSR read in the execute stage uses the registered value mstatus_fs_q, which was latched at the start of that same cycle -before the writeback fired. mstatus_fs_q only captures Dirty at the next rising clock edge, one cycle too late. 

There is existing csr_apu_stall that does not cover this case. It only fires for multi-cycle APU instructions, not LSU FP loads.

APU operations are not affected: they write back via regfile_alu_we_fw in the EX stage (one cycle earlier), so mstatus_fs_q is already updated by the time the following csrrs executes.

## Instruction sequence 
```
csrrc x0, mstatus, x11   # clear FS
csrrs x0, mstatus, x11   # set  FS ->Initial (01)
flw   f0, 0(x14)         # should transition FS -> Dirty (11)
csrrs x11, mstatus, x0   # reads FS=Initial (01) <- wrong
```
## Demonstration
A self-checking test that isolates all three cases (no NOP / 1 NOP / fadd.s) is available [here](https://github.com/karabambus/cv32e40p-dv-review/blob/feat/smf-pipeline-hazard-demo/tests/programs/custom/smf_pipeline_hazard_demo/test.c).

davideschiavone:


davideschiavone:
this PR would fix #1060

MikeOpenHWGroup (approved):
LGTM @davideschiavone

PR #1070, opened 2026-09-02
rtl: fix mstatus.FS pipeline hazard after flw

### LINKED ISSUE #1060: mstatus.FS not visible to immediately following csrrs mstatus after flw

devtyagi3909:
Fixes #1060.

As reported, a pipeline hazard occurs when reading `mstatus.FS` immediately after an `flw`. The load in the WB stage asserts `fregs_we_i` and sets `mstatus_fs_n = FS_DIRTY` combinatorially, but the `csrrs` in the EX stage reads the latched `mstatus_fs_q` one cycle too early.

This PR introduces an `mstatus_fs_fwd` wire that forwards the `FS_DIRTY` state directly into the `csr_rdata_int` generation if `fregs_we_i` or `fflags_we_i` is asserted. This ensures that the CSR read correctly observes the architectural state updates of logically older in-flight instructions (like the `flw` in WB) without incorrectly capturing the current instruction's write data.


rtl/cv32e40p_cs_registers.sv
```diff
@@ -215,6 +215,8 @@ module cv32e40p_cs_registers
   Status_t mstatus_q, mstatus_n;
   logic mstatus_we_int;
   FS_t mstatus_fs_q, mstatus_fs_n;
+  FS_t mstatus_fs_fwd;
+  assign mstatus_fs_fwd = (FPU == 1 && ZFINX == 0 && (fregs_we_i || fflags_we_i)) ? FS_DIRTY : mstatus_fs_q;
```
MikeOpenHWGroup: The coding style followed here is to first declare all variables and then express the logical functions.  Please move this `assign` statement to a line below the declarations.

This is a relatively complex expression, so please also add a one-liner comment explaining the purpose.

MikeOpenHWGroup (changes requested):

devtyagi3909:
Thanks for the review! I have updated the coding style and moved the assign block down with an inline comment as requested.

rtl/cv32e40p_cs_registers.sv
devtyagi3909: Done. Moved the assign block down below the declarations and added a quick comment explaining the WB forwarding.

MikeOpenHWGroup:
Hi @mpezzin, you might be interested in this.

MPEZZIN:
Hi @devtyagi3909,
thanks a lot for your proposed fix.
Before going further, could you please give more information on the configuration used to make your tests ?
Since the issue is caused by a race condition, I suspect your fix is only a partial fix and the issue remains in pipelined configurations of the FPU.
And please don't worry if this PR remains open for some time. There's underlying work in the non-regression test suite and we want to have it back to a working state before accepting any new contributions.

devtyagi3909:
hey matteo,

i didn't run the ACT4 tests myself - i just saw karabambus isolate the hazard in #1060 and wrote the fix for it based on that.

i looked at the pipelined FPU case and i'm pretty sure it's actually covered here. the existing `csr_apu_stall` already catches multicycle APU ops (latency >= 2) and keeps the CSR instruction stalled in ID. for latency=1 APU ops, they slip past the stall but write back at the exact cycle the csr instruction enters EX, so the forwarding logic here catches it perfectly anyway (same as it does for `flw`).

no rush on the review, completely understand you guys are fixing up the test suite right now.

